PR #37, opened 2020-03-30
Fi 722 - removing tests for non uscdi

czh-orz:
This PR removes tests for delayed sequences. It now sets encounter as a delayed resource because it's not in USCDI.  Now delayed sequences will only have the read, profile validation, and reference resolution tests. It will do them on all references found to those resources from the prior searches.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-722

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


arscan:
@czh-orz could you rebase and regenerate the tests.  Also the `onc_program_module.yml` file needs to be updated with the new order, I think?

czh-orz:
rebased and fixed

Jammjammjamm:
When I run the generator, I get different output than what is committed here. I think it is the same problem @okeefm encountered related to missing files required for generation.

lib/modules/uscore_v3.1.0/bodyheight_sequence.rb
```diff
@@ -478,9 +478,78 @@ def patient_ids
         skip 'No Provenance resources were returned from this search' unless provenance_results.present?
       end
 
-      test :validate_resources do
+      test 'All must support elements are provided in the Observation resources returned.' do
```
Jammjammjamm, line 481: Is this test reordering intended?

okeefm:
I re-added the missing files in my PR, so getting that merged would fix this.

lib/modules/uscore_v3.1.0/bodyheight_sequence.rb
czh-orz: No, I changed it back.

arscan:
> I re-added the missing files in my PR, so getting that merged would fix this.

Is this fixed @okeefm

okeefm:
Not yet, #44 is still outstanding, but I'm rebasing and fixing the changes needed now that #45 has been merged. Then it should be ready to go.

arscan:
I wonder if we need to better document why we aren't including PractitionerRole and Location (the answer: because there are no references to those US Core profiles within USCDI-mapped FHIR resources).  Because it *looks* like a gap.

I'm starting to think that what we want to do is have the US Core tests include them.  Then have an input that is simply uscdi_constrained, and if that is set by the tester it omits those tests.  And in the program edition we have the flag locked to true.

But I think we can start with this and discuss if that is the appropriate approach in the future.

PR #52, opened 2020-04-14
fi-749 better error messages when search params don't match

czh-orz:
This PR makes the error messages better when parameters don't match. Now it's in the format of:
clinical-status in  AllergyIntolerance/SMART-AllergyIntolerance-28 (["active"]) does not match clinical-status requested (["inactive"])

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-749

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1176,67 +1176,19 @@ module Inferno
       def create_search_validation(sequence)
         search_validators = ''
         sequence[:search_param_descriptions].each do |element, definition|
-          search_validators += %(
-              when '#{element}')
           type = definition[:type]
-          path_parts = definition[:path].split('.')
-          path_parts = path_parts.map { |part| part == 'class' ? 'local_class' : part }
-          path_parts.shift
-          case type
-          when 'Period', 'date'
-            search_validators += %(
-                values_found = resolve_path(resource, '#{path_parts.join('.')}')
-                match_found = values_found.any? { |date| validate_date_search(value, date) }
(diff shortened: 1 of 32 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -1122,34 +1122,37 @@ def create_search_validation(sequence)
           case type
           when 'Period', 'date'
             search_validators += %(
-                value_found = resolve_element_from_path(resource, '#{path_parts.join('.')}') { |date| validate_date_search(value, date) }
-                assert value_found.present?, '#{element} on resource does not match #{element} requested'
+                values_found = resolve_path(resource, '#{path_parts.join('.')}')
+                match_found = values_found.any? { |date| validate_date_search(value, date) }
+                assert match_found, "#{element} in #{sequence[:resource]}/\#{resource.id} (\#{values_found}) does not match #{element} requested (\#{value})"
```
okeefm, line 1127: Does it make any sense to take some/all of this matching code and move it to a function? Or at least the assertion failure message, since it's the same for pretty much all of the searches listed here.

arscan: Stylistically I think the generated code is what matters.  Does this comment apply to what is generated, or this code here?  I bring it up because I'd personally like to avoid the situation where this code is clean but then it generates very sparse tests that are just a bunch of function calls that only get called once in a given sequence (making it hard to follow).

czh-orz: I can separate the assertion failure message or the validation code in the generator do remove some of the duplication. This won't affect the generated code.

czh-orz: I updated it so that the type specific stuff is in its own function in the generator.

PR #57, opened 2020-04-15
Add a set_params call to ndjson GET requests, so they correctly validate SSL

okeefm:
The `OpenSSL::SSL::SSLContext` object in `BulkDataGroupExportValidationSequence#streamed_ndjson_get` seems like it wasn't properly set up with the right default parameters. Calling `set_params` without any options sets what OpenSSL refers to here as "sane defaults": https://docs.ruby-lang.org/en/2.4.0/OpenSSL/SSL/SSLContext.html#method-i-set_params. This seems to fix it locally (on Suspend), but I haven't tested it from infernotest or inferno.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-758

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
@@ -159,6 +159,7 @@ module Inferno
       def streamed_ndjson_get(url, headers)
         ctx = OpenSSL::SSL::SSLContext.new
         ctx.verify_mode = OpenSSL::SSL::VERIFY_PEER # set globally to VERIFY_NONE if disable_verify_peer set
+        ctx.set_params
         response = HTTP.headers(headers).get(url, ssl_context: ctx)
 
         # We need to log the request, but don't know what will be in the body


arscan:
What is the bug?  And I would think that set_params would be done first, before you update the verify mode (so it doesn't get overwritten by whatever set_params does), but I haven't looked at the docs. Just make sure you have tried it both with peer verification on and off.

radamson:
Fixes this issue in running the bulk data tests (there was a discussion in slack, but we should include that detail in the JIRA ticket):

image

Also it looks like `set_params` does things differently based on `verify_mode` so I think we'd actually want it second. [ruby 2.5.1 set_params docs](https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-set_params)

okeefm:
@arscan Yeah, what @radamson said on both counts

PR #62, opened 2020-04-22
FI-768 Validate all resources even some is invalid

yunwwang:
If one resource is invalid, instead of raising AssertionException, saving errors in an array and continue to the next resources. So the validation does not stop at the first invalid resource.
Also add logic to collect resources for response logging.
Add two more unit tests

**Submitter:**
- [x] Internal ticket for this PR: FI-768

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/views/onc_program.erb
+++ b/lib/app/views/onc_program.erb
@@ -553,7 +553,7 @@
               label: 'Limit validation to a maximum resource count',
               description: %(
                 While testing we need to validate resources in downloaded NDJSON file.
-                Please enter a integer to limit the number of resources validated per resource type.                 
+                Please enter an integer to limit the number of resources validated per resource type.                 
                 To validate all resources, leave this blank.
               ),              
               value: instance.bulk_lines_to_validate}); %>
--- a/lib/app/views/onc_program.erb
+++ b/lib/app/views/onc_program.erb
@@ -553,7 +553,7 @@
               label: 'Limit validation to a maximum resource count',
               description: %(
                 While testing we need to validate resources in downloaded NDJSON file.
-                Please enter a number to limit the number of resources validated per resource type.                 
+                Please enter a integer to limit the number of resources validated per resource type.                 
                 To validate all resources, leave this blank.
               ),              
               value: instance.bulk_lines_to_validate}); %>
--- a/lib/app/views/onc_program.erb
+++ b/lib/app/views/onc_program.erb
@@ -553,8 +553,7 @@
               label: 'Limit validation to a maximum resource count',
               description: %(
                 While testing we need to validate resources in downloaded NDJSON file.
-                Please enter a number to limit the number of resoruces validated per resource type.                 
-                To skip resource validation, enter 0.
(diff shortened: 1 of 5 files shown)


lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -120,7 +137,16 @@ def check_file_request(file, klass, validate_all, lines_to_validate, must_suppor
...
+
+          error_collection[line_count] = errors unless errors.empty?
+        end
+
+        unless error_collection.empty?
+          response_for_log[:body] = line_collection.join
+          LoggedRestClient.record_response(request_for_log, response_for_log)
+
+          index, errors = error_collection.first
+          assert false, "#{error_collection.size} / #{line_count} #{klass} resources failed profile validation. The first failed is #{klass} ##{index}: #{errors}"
```
okeefm, line 149: Just a small phrasing update
```suggestion
          assert false, "#{error_collection.size} / #{line_count} #{klass} resources failed profile validation. The first failed resource is #{klass} ##{index}: #{errors}"
```

okeefm, line 149:
```suggestion
          assert false, "#{error_collection.size} / #{line_count} #{klass} resources failed profile validation. The first failed resource is #{klass} ##{index}: \n\n#{errors[:errors].join("\n")}"
```

okeefm:
Made one suggestion for an updated error message format. Other than that, looks good to me.

arscan:
I was getting a weird empty error on Practitioner:

Screen Shot 2020-04-27 at 1 50 12 PM

And in digging into it I came to realize it was because what is called an `error` that comes back from the validator, really is an object that can include errors, warnings, and informational messages.   I updated the message to dump the whole errors object (and not errors[:error]) on Practioner to get this:

Screen Shot 2020-04-27 at 1 56 52 PM

Note no errors, but warnings and informationals.

This needs to be updated to accomodate that.  I assume we want to display at least warnings and not swallow them?  And I guess same with Information?  This nice thing about those is they are 'hidden' in the 'details' popup, so maybe we don't need to do the same rollup here.

lib/app/views/onc_program.erb
```diff
@@ -553,9 +553,9 @@
               label: 'Limit validation to a maximum resource count',
               description: %(
                 While testing we need to validate resources in downloaded NDJSON file.
-                Please enter the number of lines/resources that would be validated for each NDJSON file. 
+                Please enter a number to limit the number of resoruces validated per resource type.
```
arscan, line 556: Typo on `resources` and I don't think you need the `To skip resource validation...` line, I feel like limiting to 0 is good enough.

yunwwang: Good catch. Fixed wording.

arscan:
There are 2 requests logged for each now.  Did you add logic that by inadvertently logs the same thing two times?

Screen Shot 2020-04-28 at 3 35 12 PM

PR #74, opened 2020-05-11
Add Vital sign profile

yunwwang:
The changes includes:
Add tests for leading and trailing empty line in ndjson
Remove Location, Organization, and PractitionerRole tests from Bulk Data sequence
Add Provenance test to Bulk Data sequence

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -282,7 +282,7 @@ def streamed_ndjson_get(url, headers)
         end
 
         if line_count > MAX_RECENT_LINE_SIZE
-          response_for_log[:body] = "NOTE: RESPONSE TRUNCATED\nINFERNO ONLY DISPLAYS MOST RECENT #{MAX_RECENT_LINE_SIZE} LINES\n\n#{response_for_log[:body]}"
+          response_for_log[:body] = "NOTE: RESPONSE TRUNCATED\nINFERNO ONLY DISPLAYS FHIRST #{MAX_RECENT_LINE_SIZE} LINES\n\n#{response_for_log[:body]}"
```
arscan, line 285: FHIRst 😃 

Did you change the logger behavior so it saves the first X?  I think I had it at 'most recent' before, but i agree that given how we are referencing line numbers for errors this previous behavior made it so you couldn't figure out what line the errors referred to.

lib/app/utils/validation.rb
```diff
@@ -75,7 +75,8 @@ def self.definitions
       head_circumference: 'http://hl7.org/fhir/StructureDefinition/headcircum',
       body_weight: 'http://hl7.org/fhir/StructureDefinition/bodyweight',
       bmi: 'http://hl7.org/fhir/StructureDefinition/bmi',
-      blood_pressure: 'http://hl7.org/fhir/StructureDefinition/bp'
+      blood_pressure: 'http://hl7.org/fhir/StructureDefinition/bp',
+      vital_signs: 'http://hl7.org/fhir/StructureDefinition/vitalsigns'
```
arscan, line 79: I thought we had decided that US Core and USCDI does not require the 'generic' vital signs profile, so we do not have any regulatory or standards backing to fail them if they send a resource that looks like a vital sign by having a vital-signs coding (but do not fall into one of the profiles that are required).  Or did we find something that would give us that ability (I really doubt it because g10 is very specific about only requiring certain profiles that USCDI maps to, and vital signs isn't one of them).

lib/app/utils/validation.rb
```diff
@@ -177,6 +178,11 @@ def self.guess_r4_profile(resource)
         return DEFINITIONS[US_CORE_R4_URIS[:heart_rate]] if observation_contains_code(resource, '8867-4')
 
         return DEFINITIONS[US_CORE_R4_URIS[:body_temperature]] if observation_contains_code(resource, '8310-5')
+
+        return DEFINITIONS[US_CORE_R4_URIS[:vital_signs]] if resource&.category&.first&.coding&.any? { |coding| coding&.code == 'vital-signs' }
```
arscan, line 182: If we do keep vital signs (which I feel like we shouldn't without very strong proof), we should look in any category here, not just the first.  something like `resource&.category&.any{|category| category&.coding&.any{|coding| coding... } }`

lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -114,16 +114,16 @@ def check_file_request(file, klass, validate_all, lines_to_validate, must_suppor
...
             resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class, p.url)
-
-            # Remove warnings if using internal FHIRModelsValidator. FHIRModelsValidator has an issue with FluentPath.
-            resource_validation_errors = [] if resource_validation_errors[:errors].empty? &&
-                                               (Inferno::RESOURCE_VALIDATOR.is_a?(Inferno::FHIRModelsValidator) ||
-                                               (resource_validation_errors[:warnings].empty? && resource_validation_errors[:information].empty?))
           else
             warn { assert false, 'No profiles found for this Resource' }
-            resource_validation_errors = resource.validate
+            resource_validation_errors = Inferno::RESOURCE_VALIDATOR.validate(resource, versioned_resource_class)
```
arscan, line 119: Great catch!

lib/app/utils/validation.rb
```diff
@@ -177,6 +178,11 @@ def self.guess_r4_profile(resource)
...
+
+        return nil
```
arscan, line 184: I think it makes more sense to just let if `fall through` to the end and have it handled in the same way as `DiagnosticReport`, at the end of the method.  Right now, if there are no matching Observation profiles you return `nil`, but if there are no matching `DiagnosticReport` you are returning a random DiagnosticReport profile (with that `candidates.first` thing, which is just wrong).  Also, are we sure that all the code everywhere that calls this method can handle a `nil` being returned instead of a profile?

lib/modules/onc_program/bulk_data_group_validation_sequence.rb
yunwwang: Sometime, I cannot even spell the word "fire" correctly. Yes I did change the order of logs kept. Typo fixed

lib/app/utils/validation.rb
yunwwang: Yes. This is a valid concern. So another way to address this is adding additional logic in bulk data sequence.

lib/app/utils/validation.rb
yunwwang: fixed

lib/app/utils/validation.rb
yunwwang: As previous comments. Logic for vital-signs are moved to bulk data validation sequence. Added a keyword parameter use_default to direct guess_profile() to use the default profile or to return nil

arscan:
I don't really understand why we are adding special code here for bulk data to pass in different parameters to guess_profile.  This seems to be adding complexity to something that is already probably too complex.  If I looked at the method signature 'guess_profile', i would have no idea what "use_default" parameter means unless I looked at the code.  In looking at the code, i'd call it "randomly_pick_a_profile_from_those_of_same_type_resource_if_no_suitable_profiles_found" which is an obviously wrong thing to do :). I really think we need to fix this instead of adding a flag to work around it on one side of our codebase.

Since I feel like this needs soem discussion internally so we can align us core queries and bulk data profile resolution, can we isolate the other parts of this PR (Add tests for leading and trailing empty line in ndjson, Remove Location, Organization, and PractitionerRole tests from Bulk Data sequence, Add Provenance test to Bulk Data sequence) and get those in because those are pretty straightforward, and address this guess_profile thing separately.

yunwwang:
Changes for vital-sign profile are moved to another branch. This PR contains:
Add tests for leading and trailing empty line in ndjson
Remove Location, Organization, and PractitionerRole tests from Bulk Data sequence
Add Provenance test to Bulk Data sequence

PR #94, opened 2020-05-27
FI-834 bulk data terminology validation

yunwwang:
This pull request contains changes to 
1) added terminology binding validation for bulk data
2) removed default parameters values from some unit tests

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -920,49 +920,6 @@ module Inferno
         end
       end
 
-      def validate_bindings(bindings, resources)
-        return unless bindings.present?
-
-        invalid_binding_messages = []
-        invalid_binding_resources = Set.new
-        validation_errors = { errors: [], warnings: [] }
-        bindings.select { |binding_def| binding_def[:strength] == 'required' }.each do |binding_def|
-          begin
-            invalid_bindings = resources_with_invalid_binding(binding_def, resources)
-          rescue Inferno::Terminology::UnknownValueSetException => e
-            validation_errors[:warnings] << e.message
-            invalid_bindings = []
-          end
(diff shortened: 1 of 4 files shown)


lib/app/sequence_base.rb
```diff
@@ -920,6 +920,49 @@ def find_slice_by_values(element, values)
         end
       end
 
+      def validate_bindings(bindings, resources)
```
arscan, line 923: I know `sequence_base` has traditionally been a dumping ground for shared methods but we have to stop doing this.  We need to be moving in the direction of `sequence_base` purely being 'Inferno DSL' definitions (e.g. `test`, `omit`, `metadata`), otherwise we will never be able to separate the 'platform' from the 'tests'.

It looks like this method is only used in the bulk_data_group_validation sequence, so can we just move it there and not touch sequence_base?

yunwwang: This method can be used by single patient tests. I would like to create a separate ticket to update test generator to use this method for terminology validation

arscan: If you are going to do a bit of refactor to get the single patient query API tests sharing the same code in another ticket, lets just keep this method out of sequence base and then move it to that shared place as part of that work (almost definitely not into sequence_base but that other shared/utility/helper module, along with the dozen+ other us-core specific methods we have in there).  We need to move in the right direction from here on out.

yunwwang: The function is moved to bulk_data_group_validation_sequence.

czh-orz:
I think we're ignoring the warnings from the binding validation. Is that intentional?

yunwwang:
When using FHIR Model validator, MUST_SUPPORT and Terminology binding validation are not performed.

PR #98, opened 2020-05-29
fi-840 Fix error when cannot find codeableconcept search value

czh-orz:
This fixes an issue where we give an error when trying to perform a search with system for a codeableconcept that cannot be found. 

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-840
- [ ] Tests/code quality metrics have been run locally and pass

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -808,7 +808,7 @@ module Inferno
                        when FHIR::CodeableConcept
                          if include_system
                            coding_with_code = resolve_element_from_path(element, 'coding') { |coding| coding.code.present? }
-                           "#{coding_with_code.system}|#{coding_with_code.code}"
+                           coding_with_code.present? ? "#{coding_with_code.system}|#{coding_with_code.code}" : nil
                          else
                            resolve_element_from_path(element, 'coding.code')
                          end


lib/app/sequence_base.rb
```diff
@@ -808,7 +808,7 @@ def get_value_for_search_param(element, include_system = false)
                        when FHIR::CodeableConcept
                          if include_system
                            coding_with_code = resolve_element_from_path(element, 'coding') { |coding| coding.code.present? }
-                           "#{coding_with_code.system}|#{coding_with_code.code}"
+                           coding_with_code.present? ? "#{coding_with_code.system}|#{coding_with_code.code}" : nil
```
Jammjammjamm, line 811: Is this nil-safe everywhere? As an example, I see nil checks here: https://github.com/onc-healthit/inferno-program/blob/development/lib/modules/uscore_v3.1.0/us_core_careplan_sequence.rb#L222
but not here: https://github.com/onc-healthit/inferno-program/blob/development/lib/modules/uscore_v3.1.0/us_core_careplan_sequence.rb#L184

czh-orz: We don't explicitly check for nil in the second example, but the validate_reply_entries acts as a nil check since we won't reach that line if there was no category in the resource.

PR #105, opened 2020-06-01
FI-854 Add standalone launch with public client.

arscan:
Adds public launch requirement to the 'other' tests.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/onc_program/onc_parameters.rb
+++ b/lib/modules/onc_program/onc_parameters.rb
@@ -15,6 +15,11 @@ module Inferno
       property :onc_sl_patient_id, String
       property :onc_sl_oauth_token_endpoint, String
 
+      property :onc_public_confidential_client, Boolean
+      property :onc_public_client_id, String
+      property :onc_public_client_secret, String
+      property :onc_public_scopes, String
+
       property :onc_patient_ids, String
 
       property :onc_visual_single_registration, String, default: 'false'
--- a/lib/modules/onc_program/onc_parameters.rb
+++ b/lib/modules/onc_program/onc_parameters.rb
@@ -13,6 +13,7 @@ module Inferno
       property :onc_sl_token, String
       property :onc_sl_refresh_token, String
       property :onc_sl_patient_id, String
+      property :onc_sl_oauth_authorize_endpoint, String
       property :onc_sl_oauth_token_endpoint, String
 
       property :onc_public_client_id, String
--- a/lib/modules/onc_program/onc_parameters.rb
+++ b/lib/modules/onc_program/onc_parameters.rb
@@ -15,9 +15,7 @@ module Inferno
       property :onc_sl_patient_id, String
       property :onc_sl_oauth_token_endpoint, String
 
-      property :onc_public_confidential_client, Boolean
       property :onc_public_client_id, String
-      property :onc_public_client_secret, String
       property :onc_public_scopes, String
 
       property :onc_patient_ids, String
--- a/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
+++ b/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
@@ -14,8 +14,6 @@ module Inferno
       test_id_prefix 'OSLSP'
 
       requires :onc_public_client_id,
-               :onc_public_confidential_client,
-               :onc_public_client_secret,
                :onc_public_scopes,
                :onc_sl_oauth_authorize_endpoint,
                :onc_sl_oauth_token_endpoint,
--- a/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
+++ b/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
@@ -171,9 +171,7 @@ module Inferno
           versions :r4
         end
         assert @token_response_body.key?('id_token'), 'Token response did not provide an id_token as required.'
-
       end
-
     end
   end
 end
--- a/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
+++ b/lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
@@ -7,7 +7,7 @@ module Inferno
     class OncStandalonePublicLaunchSequence < SequenceBase
       include Inferno::Sequence::SharedONCLaunchTests
 
-      title 'Public Client Standalone Launch Sequence'
+      title 'Public Client Standalone Launch with OpenID Connect'
 
       description 'Register Inferno as a public client with patient access and execute standalone launch.'
 
@@ -53,7 +53,7 @@ module Inferno
       end
 
       def required_scopes
-        ['launch/patient']
+        ['launch/patient', 'openid', 'fhirUser']
       end
 
       details %(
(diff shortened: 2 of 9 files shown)


lib/modules/onc_program_module.yml
```diff
@@ -172,18 +209,18 @@ test_sets:
         run_all: false
         run_skipped: false
         overview: >
-          Not all requirements that need to be tested fit within the previous scenarios.  The tests contained in this section addresses remaining testing requirements.
-          Each of these tests need to be run independently.  Please read the
-          instructions for each, as they require special attention on the part of the tester.
-
-          Note: additional tests are expected to be added to this preview version of Inferno prior to rule effective date.
+          Not all requirements that need to be tested fit within the previous
+          scenarios.  The tests contained in this section addresses remaining
```
Jammjammjamm, line 213:
```suggestion
          scenarios.  The tests contained in this section address remaining
```

lib/modules/onc_program_module.yml
```diff
@@ -172,18 +209,18 @@ test_sets:
...
+          testing requirements. Each of these tests need to be run
+          independently.  Please read the instructions for each in the 'About'
+          section, as they may special setup on the part of the tester.
```
Jammjammjamm, line 216:
```suggestion
          section, as they may require special setup on the part of the tester.
```

lib/modules/onc_program/onc_parameters.rb
```diff
@@ -15,6 +15,11 @@ class TestingInstance
       property :onc_sl_patient_id, String
       property :onc_sl_oauth_token_endpoint, String
 
+      property :onc_public_confidential_client, Boolean
+      property :onc_public_client_id, String
+      property :onc_public_client_secret, String
```
radamson, line 20: What is this used for?  I don't see it as a field in the "run" modal and it doesn't look like it's used anywhere.

arscan: You are correct, we just need the client_id.  I added those because i thought they would have to exist to have the sequence work, but then realized they didn't have to (but then forgot to remove them here).  Removing...

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
```diff
@@ -0,0 +1,188 @@
...
+
+      def instance_client_id
+        @instance.onc_public_client_id
+      end
+
+      def instance_confidential_client
+        false
+      end
+
+      def instance_client_secret
```
radamson, line 99: I don't think this method is necessary and it seems weird to declare this in a public launch file (I understand we may need to for some shared launch tests).  I looked in the shared_onc_launch tests and all the references to this were wrapped in a check for instance_confidential_client so I think it's safe to delete.

arscan: I did it just to be thorough and to be clear that this test won't be using a client_secret.  It defaults to the client secret used by the EHR launch, and it seems risky to just assume that we have enough guards to not send that along. I think this is ok to leave.

As is probably pretty clear, we have a good opportunity here to make what I"m trying to do much more straightforward than it is.  So at a certain point, I'm just accepting that this is hard to read, but functional, and we can consider what this is doing for future designs.

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
```diff
@@ -0,0 +1,188 @@
...
+      title 'Public Client Standalone Launch Sequence'
+
+      description 'Register Inferno as a public client with patient access and execute standalone launch.'
+
+      test_id_prefix 'OSLSP'
+
+      requires :onc_public_client_id,
+               :onc_public_scopes,
+               :onc_sl_oauth_authorize_endpoint,
+               :onc_sl_oauth_token_endpoint,
```
radamson, line 19: Is there any reason this sequence uses `onc_sl_oauth_token_endpoint` (defined by `onc_standalone token_refresh_sequence`) while `onc_standalone_launch_sequence` uses `oauth_token_endpoint` defined by `onc_smart_dsicovery_sequence`?

`oauth_token_endpoint` is nice because you can keep testing even if the token refresh fails and they should be the same anyways.

https://github.com/onc-healthit/inferno-program/blob/6ec57461a2e9d490f36965261c47a3fe14109053/lib/modules/onc_program/onc_standalone_token_refresh_sequence.rb#L62

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
```diff
@@ -0,0 +1,188 @@
...
+
+      title 'Public Client Standalone Launch Sequence'
+
+      description 'Register Inferno as a public client with patient access and execute standalone launch.'
+
+      test_id_prefix 'OSLSP'
+
+      requires :onc_public_client_id,
+               :onc_public_scopes,
+               :onc_sl_oauth_authorize_endpoint,
```
radamson, line 18: This doesn't display because there is only a field for `oauth_authorize_endpoint`.

I think you want to use `oauth_authorize_endpoint` here anyway for similar reasons as `oauth_token_endpoint`
Screen Shot 2020-06-01 at 3 51 48 PM

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
radamson: I don't agree that it makes it clear that it won't be using a client_secret (if anything it makes me suspicious that it is!), but I understand the concern about guards.  Although, if those guards are not properly instituted this might hide other issues (like an `Authorization` header being set).

Seeing as we don't have much time to fix this I'm happy with making this more straightforward and logical at a later date.

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
arscan: This is where things get difficult because I can't chain a discovery sequence in front of this or the token revocation sequences.

EHR launches could use a different auth server.  As we run through the normal flow of the app, the auth server picked up in the standalone launch discovery will get overwritten by the auth server in the ehr launch discovery.  Then we will prefill with a wrong auth server.  This is a bigger deal on token revocation, where we need to be revoking the standalone patient token and not the ehr token, but by default when walking through we will prefill with the ehr auth server.

So, just to make this work, the standalone discovery launch steps saves the auth server 'off to the side' as a general record that the token revocation sequence and this sequence can use.

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
radamson: Also later in the test (line 129) `oauth_token_endpoint` is referenced, so it actually is the one that is required.

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
```diff
@@ -0,0 +1,188 @@
...
+          'Observation',
+          'Organization',
+          'Practitioner',
+          'PractitionerRole',
+          'Procedure',
+          'Provenance'
+        ]
+      end
+
+      def required_scopes
```
radamson, line 55: This doesn't appear to be ever used.  `onc_standalone_launch_tests` use this with `required_scope_test` from the `shared_onc_launch_tests`.  I'm not sure you'd want to use that test out of the box though because the title says "access with OpenID Connect and Refresh Token scopes used" and those scopes aren't actually checked.

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
```diff
@@ -0,0 +1,188 @@
...
+
+      invalid_client_id_test(index: '05')
+
+      successful_token_exchange_test(index: '06')
+
+      token_response_contents_test(index: '07')
+
+      token_response_headers_test(index: '08')
+
+      test :unauthorized_read do
```
radamson, line 162: It's probably not the right time to do this, but this should be in the `shared_onc_launch_tests` for DRYness.

lib/modules/onc_program_module.yml
```diff
@@ -12,14 +12,22 @@ test_sets:
         short_label: Standalone Patient App
         run_all: true
         overview: |
-          This scenario demonstrates the ability to perform a Patient Standalone Launch to a [SMART on FHIR](http://www.hl7.org/fhir/smart-app-launch/) client with a patient context,
-          refresh token, and [OpenID Connect (OIDC)](https://openid.net/specs/openid-connect-core-1_0.html) identity token.  After launch, a simple Patient resource read is performed
-          on the patient in context.  The access token is then refreshed, and the Patient resource is read using the new
-          access token to ensure that the refresh was successful. The authentication information provided by OpenID Connect is decoded and validated, and simple queries
-          are performed to ensure that access is granted to all USCDI data elements.
+          This scenario demonstrates the ability to perform a Patient Standalone
```
radamson, line 15:
```suggestion
          This scenario demonstrates the ability of a system to perform a Patient Standalone
```

lib/modules/onc_program/onc_standalone_public_launch_sequence.rb
arscan: Ah, yes, I failed to realize that the authorize endpoint was different than the token endpoint.  But unfortunately, it cannot be simply 'oauth_authorize_endpoint' because that will be prefilled by the EHR launches info, which I don't want.  Reworking now...

PR #117, opened 2020-06-10
Fix ordering of RequestResponses

Jammjammjamm:
RequestResponses were being shown out of order due to insufficient precision in how timestamps are stored in the database. This branch adds a new field to store their order, as shown in the screenshots below (check the order from bottom to top).

Before:
Screen Shot 2020-06-10 at 12 52 15 PM

After:
Screen Shot 2020-06-10 at 2 45 12 PM

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/models/request_response.rb
+++ b/lib/app/models/request_response.rb
@@ -14,7 +14,7 @@ module Inferno
       property :response_body, Text
       property :direction, String
       property :instance_id, String
-      property :sequence, Serial, unique_index: true
+      property :request_index, Serial, unique_index: true
 
       property :timestamp, DateTime, default: proc { DateTime.now }
 
--- a/lib/app/models/request_response.rb
+++ b/lib/app/models/request_response.rb
@@ -14,6 +14,7 @@ module Inferno
       property :response_body, Text
       property :direction, String
       property :instance_id, String
+      property :sequence, Serial, unique_index: true
 
       property :timestamp, DateTime, default: proc { DateTime.now }
 
--- a/lib/app/models/test_result.rb
+++ b/lib/app/models/test_result.rb
@@ -27,7 +27,7 @@ module Inferno
       property :redirect_to_url, String
       property :expect_redirect_failure, Boolean, default: false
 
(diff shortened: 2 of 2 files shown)


okeefm:
To anyone testing this: if you get a `DataObjects::SyntaxError: Cannot add a PRIMARY KEY column`, the solution is to remove the `*.db` files in `<inferno root>/data/`.

Side note: we'll want to figure out how to do this automatically (or tell people to do this) if we merge this, because otherwise everyone's Inferno installs will stop booting when they upgrade to this commit.

lib/app/models/request_response.rb
```diff
@@ -14,6 +14,7 @@ class RequestResponse
       property :response_body, Text
       property :direction, String
       property :instance_id, String
+      property :sequence, Serial, unique_index: true
```
radamson, line 17: No biggie, but I think something like `request_order` would be a better name.  "Sequence" already has a different meaning in inferno and when I first saw this while inspecting the database my mind was expecting something different.

radamson:
Tested it out and it worked well!  Happy to have this merged as is or implement the name change if others agree it is a good idea.

Screen Shot 2020-06-25 at 7 35 19 AM

lib/app/models/request_response.rb
okeefm: Yeah, `request_order` or `request_index` might make more sense for this, given that Sequence is already taken.

arscan: I third this suggestion, and updated the column.  Could some verify my update real quick?

arscan:
Our plan regarding the DB change is to have users manually move or delete the existing database file.  This is a headache I wanted to avoid, but adding the complexity of migration scripts is risky without a good amount of thought.  And we need this bug (wrong ordered responses) fixed ASAP because it is misleading.

lib/app/models/request_response.rb
radamson: Tested out the change and it worked as expected 👍

PR #119, opened 2020-06-15
FI-870 Ensure correct test instance found in smart redirect when no state.

arscan:
If during the launch process you visit Infenro's launch URI without a state, inferno will match you to the wrong instance. It should see that no state param was passed, then check if there is a cookie with the instance stored there, and redirect you to the instance in the cookie.

Previous to this fix, the logic is set up so that Inferno will find you the first instance with a state of 'nil', which will be present in any instance that hasn't yet run any inferno launches.

To recreate, simply go to https://inferno.healthit.gov/inferno/oauth2/static/redirect.
**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/utils/oauth2_error_messages.rb
+++ b/lib/app/utils/oauth2_error_messages.rb
@@ -7,11 +7,7 @@ module Inferno
         %(
           <p>
             Inferno has detected an issue with the SMART launch.
-            <% if params[:state].nil?} %>
-              No 'state' parameter was returned by the authorization server.
-            <% else %>
-              No actively running launch sequences found with a 'state' parameter of '#{params[:state]}'.
-            <% end %>
+            #{param_description}
             The authorization server is not returning the correct state variable and
             therefore Inferno cannot identify which server is currently under test.
             Please click your browser's "Back" button to return to Inferno,
@@ -20,6 +16,12 @@ module Inferno
           #{server_error_message}
           #{server_error_description}
         )
+      end
+
+      def param_description
+        return 'No state parameter was returned by the authorization server. ' if params[:state].nil?
+
+        "No actively running launch sequences found with a 'state' parameter of '#{params[:state]}'."
       end
 
       def server_error_message
--- a/lib/app/utils/oauth2_error_messages.rb
+++ b/lib/app/utils/oauth2_error_messages.rb
@@ -19,7 +19,7 @@ module Inferno
       end
 
       def param_description
-        return 'No state parameter was returned by the authorization server. ' if params[:state].nil?
+        return "No 'state' parameter was returned by the authorization server." if params[:state].nil?
 
         "No actively running launch sequences found with a 'state' parameter of '#{params[:state]}'."
       end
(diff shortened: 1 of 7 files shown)


okeefm:
Is this the expected result when no state is passed in, either in the URL or a cookie?
Screen Shot 2020-06-24 at 7 57 59 AM

arscan:
Yes.  I suppose I could add a little logic in there to simply say that no state was passed.

arscan:
@okeefm Added a test that highlights the issue, and cleared out some DB tables to ensure these integration tests have more of a clean slate for the parts of the DB that matter.  Also cleaned up the language a bit when state is blank.  Could you just take a quick revisit before I merge?

lib/app/utils/oauth2_error_messages.rb
```diff
@@ -7,7 +7,11 @@ def no_instance_for_state_error_message
         %(
           <p>
             Inferno has detected an issue with the SMART launch.
-            No actively running launch sequences found with a state of #{params[:state]}.
+            <% if params[:state].nil?} %>
```
okeefm, line 10: These ERB tags are coming through in the HTML output, because (I think) this is a raw string.
Screen Shot 2020-06-25 at 4 11 20 PM

arscan: Fixed in the last commit.  Great catch.

okeefm: Yup, looks good.

lib/app/utils/oauth2_error_messages.rb
```diff
@@ -18,6 +18,12 @@ def no_instance_for_state_error_message
         )
       end
 
+      def param_description
+        return 'No state parameter was returned by the authorization server. ' if params[:state].nil?
```
radamson, line 22: `state` is wrapped in single quotes below so we should be consistent one way or another.
```suggestion
        return "No 'state' parameter was returned by the authorization server." if params[:state].nil?
```

lib/app/views/guided.erb
```diff
@@ -75,7 +75,7 @@
           No actively running launch sequences found for iss <%=params[:iss]%>.
           Please ensure that the EHR launch test is actively running before attempting to launch Inferno from the EHR.
         <% elsif error_code == 'no_state' %>
-          Inferno has detected an issue with the SMART launch. No actively running launch sequences found with a state of <%=params[:state]%>.
+          Inferno has detected an issue with the SMART launch. No actively running launch sequences found with a state parameter of '<%=params[:state]%>'.
```
radamson, line 78: it would be nice to have the same logic here that lets the user no state parameter was given.

e.g. here I gave no state parameter but Inferno was saying one was found with value of ''

Screen Shot 2020-06-26 at 9 53 01 AM

arscan: Valid suggestion, but I'm going to leave this as-is.

PR #121, opened 2020-06-18
subtract/add a day when doing searches based on period

czh-orz:
This PR fixes a bug where we incorrectly get the search value for periods that start and end on the same day. To get around this, I'm just going to subtract one day from the initial start date and add one day to the end date.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-878

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/test/unit/sequence_base_test.rb
+++ b/test/unit/sequence_base_test.rb
@@ -123,12 +123,12 @@ class SequenceBaseTest < MiniTest::Test
     end
 
     it 'returns value from period' do
-      { start: 'now', end: 'later' }.each do |key, value|
+      { start: '2020-05-26', end: '2020-05-26' }.each do |key, value|
         element = FHIR::Period.new(key => value)
         expected_value = if key == :start
-                           'gt' + value
+                           'gt2020-05-25T00:00:00+00:00'
                          else
-                           'lt' + value
+                           'lt2020-05-27T00:00:00+00:00'
                          end
         assert @sequence.get_value_for_search_param(element) == expected_value
       end
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -799,9 +799,9 @@ module Inferno
         search_value = case element
                        when FHIR::Period
                          if element.start.present?
-                           'gt' + element.start
+                           'gt' + (DateTime.xmlschema(element.start) - 1).xmlschema
                          else
-                           'lt' + element.end
+                           'lt' + (DateTime.xmlschema(element.end) + 1).xmlschema
                          end
                        when FHIR::Reference
                          element.reference


hershilpatel:
Does this add/subtract start/end date in all cases or just when the start and end date are the same? It seems like another test to test the normal case of different dates could be useful as well to make sure it does the correct behavior in that case.

czh-orz:
> Does this add/subtract start/end date in all cases or just when the start and end date are the same? It seems like another test to test the normal case of different dates could be useful as well to make sure it does the correct behavior in that case.

The add/subtract gets applied to all period searches. I don't think more tests are necessary because we just need to make sure that the new searches will definitely include the period we base the search off of. Adding/subtracting a day will ensure that.

test/unit/sequence_base_test.rb
```diff
@@ -123,12 +123,12 @@ def setup
     end
 
     it 'returns value from period' do
-      { start: 'now', end: 'later' }.each do |key, value|
+      { start: '2020-05-26', end: '2020-05-26' }.each do |key, value|
         element = FHIR::Period.new(key => value)
         expected_value = if key == :start
-                           'gt' + value
+                           'gt2020-05-25T00:00:00+00:00'
```
yunwwang, line 129: This tests are not correct. 1) you don't need to covert that to dateTime, server shall understand gt2020-05-25 2) gt2020-05-25 equals to gt2020-05-25T23:59:59.0000 (means excluding any time in that date)

test/unit/sequence_base_test.rb
```diff
@@ -123,12 +123,12 @@ def setup
...
                          else
-                           'lt' + value
+                           'lt2020-05-27T00:00:00+00:00'
```
yunwwang, line 131: Similar as above. lt2020-05-27 should be good. Actually I think this would be a good test to see if EHR implement period comparison correctly.

yunwwang:
This method  `def date_comparator_value(comparator, date)` (at sequence_base.rb line 838)   does not distinguish le from lt (and ge from gt).

test/unit/sequence_base_test.rb
czh-orz: Does this matter? All the value needs to do is make sure is that the resource we are searching by is included in the search. In this way, gt2020-05-25 is the same as gt2020-05-25T00:00:00+00:00. I can change it to gt2020-05-25, but then we aren't testing if they can accept datetime. Should I maybe just use datetime when the original value is datetime and date when the original value is a date?

yunwwang: Yes. It doesn't matter since Inferno only need to check if the resource returned.

czh-orz:
> This method `def date_comparator_value(comparator, date)` (at sequence_base.rb line 838) does not distinguish le from lt (and ge from gt). Using number comparison as example. Assume I have integer 3. "lt 3.05" and "le 3.05" both yield the correct result. But I think the 2nd one should be "le 3" to test the "eq" part

I tried this and it seems to work/not break anything. I'm just nervous about making this change now since I'm not sure if there are cases that I'm not thinking about. The "eq" part should be tested anyways in the non-comparator search.

PR #124, opened 2020-06-19
Fi 884 scrub presets

arscan:
This cleans up the presets for when you load up inferno on localhost.  It also fixes two links to the wrong github repo on the bottom of the interface.

**Submitter:**
- [x] Internal ticket for this PR: Fi 884

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -94,39 +94,17 @@ bulk_data_jwks:
     
 # preset fhir servers: optional. Minimally requires name, uri, module, optional inferno_uri, client_id, client_secret, scopes, instructions link
 presets:
-  reference_server:
-    name: FHIR Reference Server
-    uri: http://localhost:8080/r4
-    module: onc_program
+  localhost_reference_server:
     inferno_uri: http://localhost:4567
+    name: Inferno Reference Server
+    uri: https://inferno.healthit.gov/reference-server/r4
+    module: onc_program
     client_id: SAMPLE_CONFIDENTIAL_CLIENT_ID
     confidential_client: true
     client_secret: SAMPLE_CONFIDENTIAL_CLIENT_SECRET
-    onc_sl_url: http://localhost:8080/r4
+    onc_sl_url: https://inferno.healthit.gov/reference-server/r4
     onc_sl_client_id: SAMPLE_CONFIDENTIAL_CLIENT_ID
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -94,39 +94,17 @@ bulk_data_jwks:
     
 # preset fhir servers: optional. Minimally requires name, uri, module, optional inferno_uri, client_id, client_secret, scopes, instructions link
 presets:
-  reference_server:
-    name: FHIR Reference Server
-    uri: http://localhost:8080/r4
-    module: onc_program
+  localhost_reference_server:
```
radamson, line 97: This preset was renamed to `localhost_reference_server` but it points to the reference server on `inferno.healthit.gov`.  Is this intentional? seems a little confusing...
```suggestion
  inferno_health_reference_server:
```

lib/app/views/layout.erb
```diff
@@ -62,10 +62,10 @@
     <footer class="footer">
       <div class="container row">
         <div class="col-sm text-left">
-          <a target="_blank" href="https://github.com/onc-healthit/inferno">Open Source</a>
+          <a target="_blank" href="https://github.com/onc-healthit/inferno-program">Open Source</a>
         </div>
         <div class="col-sm text-center">
-          <a target="_blank" href="https://github.com/onc-healthit/inferno/issues">Issues</a>
+          <a target="_blank" href="https://github.com/onc-healthit/inferno-program/issues">Issues</a>
```
radamson, line 68: This also needs to be fixed here:

https://github.com/onc-healthit/inferno-program/blob/d941b150e2289843f1ab1401917773c95dddde90/lib/app/views/index.erb#L66

config.yml
arscan: I named it this because the most important aspect of this entry is that it only shows up when running under `localhost`.  The second most important aspect is the server it is pointing to, but instead of having two host names in the key, I just state that it is an entry to the 'reference server'.

I'm not thrilled about this implementation (why do we need a key here?), but it works for now.

lib/app/views/layout.erb
arscan: We use a different index page for inferno_program.  This only shows up for community.  Confusing, I know.

PR #132, opened 2020-06-29
FI-891: Token Exchange Unrequested Scopes

radamson:
Fix checks for unrequested scopes to only apply to token refresh.  This check was erroneously applied to the initial OAuth 2.0 token exchange.

[JIRA Issue](https://oncprojectracking.healthit.gov/support/browse/FI-891)
Relevant GH Issue: https://github.com/onc-healthit/inferno/issues/464

**Submitter:**
- [ ] Internal ticket has a justification for its Community/Program label

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/onc_program/shared_onc_launch_tests.rb
+++ b/lib/modules/onc_program/shared_onc_launch_tests.rb
@@ -45,7 +45,7 @@ module Inferno
         # with the one used in the ehr launch.
       end
 
-      def validate_token_response_contents(token_response, require_expires_in:)
+      def validate_token_response_contents(token_response, require_expires_in:, check_unrequested_scopes: false)
         skip_if token_response.blank?, no_token_response_message
 
         assert_valid_json(token_response.body)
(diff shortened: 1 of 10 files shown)


lib/modules/smart/shared_launch_tests.rb
```diff
@@ -76,8 +76,13 @@ def validate_token_response_contents(token_response, require_expires_in:)
           assert missing_scopes.empty?, "Token exchange response did not include expected scopes: #{missing_scopes}"
         end
 
-        extra_scopes = actual_scopes - expected_scopes
-        assert extra_scopes.empty?, "Token response contained unrequested scopes: #{extra_scopes.join(', ')}"
+        # During a token refresh scopes provided must be a strict sub-set of the scopes granted in the original launch.
+        # This does not apply to the original token exchange
+        # See: https://github.com/onc-healthit/inferno/issues/464 and other related issues tagged linked to that issue.
+        if check_unrequested_scopes
+          extra_scopes = actual_scopes - expected_scopes
```
arscan, line 83: I think `expected_scopes` probably should be `@instance.received_scopes`, because `expected_scopes` here is what was requested in the original request I think, not what was actually granted (received).  So if you have a situation where in the initial authorization request, the app requests `patient/*.read`, the server grants `patient/Condition.read` and `patient/Patient.read`.  On the token refresh, I'd expect it to be a subset of `patient/Condition.read` and `patient/Patient.read`, not `patient/*.read`, no?

This is complicated because of how the code is shared though, so we need to be careful here.

radamson: Good catch.  Looking into a fix now, but I'm wondering if we'll also have issues if we are granted something with a wildcard originally and then a subset later

lib/modules/onc_program/shared_onc_launch_tests.rb
```diff
@@ -101,8 +101,13 @@ def validate_token_response_contents(token_response, require_expires_in:)
...
         end
 
-        extra_scopes = actual_scopes - expected_scopes
-        assert extra_scopes.empty?, "Token response contained unrequested scopes: #{extra_scopes.join(', ')}"
+        # During a token refresh scopes provided must be a strict sub-set of the scopes granted in the original launch.
+        # This does not apply to the original token exchange
+        # See: https://github.com/onc-healthit/inferno/issues/464 and other related issues tagged linked to that issue.
+        if check_scope_subset
+          extra_scopes = actual_scopes - @instance.received_scopes.split(' ')
+          assert extra_scopes.empty?, "Token response contained scopes which are not a subset of those provided in the original launch: #{extra_scopes.join(', ')}"
```
arscan, line 109:
```suggestion
          assert extra_scopes.empty?, "Token response contained scope which is not a subset of the scope granted to the original access token: #{extra_scopes.join(', ')}"
```

PR #135, opened 2020-07-01
fix period comparator search

czh-orz:
This PR fixes a bug where lt and le comparator searches will search by the start date instead of the day after the start date of periods. This is caused by a recent change to how the normal period searches are performed. 

To test, the procedure sequence should pass with the reference data set.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-906
- [ ] Tests/code quality metrics have been run locally and pass

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -841,9 +841,9 @@ module Inferno
 
         case comparator
         when 'lt', 'le'
-          comparator + (DateTime.xmlschema(date) + 1).xmlschema
+          comparator + (DateTime.xmlschema(date) + 2).xmlschema
         when 'gt', 'ge'
-          comparator + (DateTime.xmlschema(date) - 1).xmlschema
+          comparator + (DateTime.xmlschema(date) - 2).xmlschema
         else
           ''
         end
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -837,13 +837,12 @@ module Inferno
       end
 
       def date_comparator_value(comparator, date)
-        date = date.slice(2..-1) if ['gt', 'ge', 'lt', 'le'].include? date[0, 2]
-
+        date = date.start || date.end if date.is_a? FHIR::Period
         case comparator
         when 'lt', 'le'
-          comparator + (DateTime.xmlschema(date) + 2).xmlschema
+          comparator + (DateTime.xmlschema(date) + 1).xmlschema
         when 'gt', 'ge'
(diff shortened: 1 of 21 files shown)


lib/app/sequence_base.rb
```diff
@@ -841,9 +841,9 @@ def date_comparator_value(comparator, date)
 
         case comparator
         when 'lt', 'le'
-          comparator + (DateTime.xmlschema(date) + 1).xmlschema
+          comparator + (DateTime.xmlschema(date) + 2).xmlschema
```
yunwwang, line 844: Why adding 2?

radamson: @czh-orz just explained this to me so let me see if I can explain it right...

Search Parameters for searches are first calculated and a search is performed with them:

e.g.
https://github.com/onc-healthit/inferno-program/blob/cda7ed0acf3a2de8ec2799bca08903964374f54f/lib/modules/uscore_v3.1.0/us_core_procedure_sequence.rb#L222-L225

Then the search parameters for the comparator search are calculated _based on the previous search parameter value_
e.g.
https://github.com/onc-healthit/inferno-program/blob/cda7ed0acf3a2de8ec2799bca08903964374f54f/lib/modules/uscore_v3.1.0/us_core_procedure_sequence.rb#L238

When we do the initial search for a Period `get_value_for_search_param` decrements the date because a start exists
https://github.com/onc-healthit/inferno-program/blob/cda7ed0acf3a2de8ec2799bca08903964374f54f/lib/app/sequence_base.rb#L800-L802

Then as we loop through each of the comparators based on this date any of the `lt` or `le` options will increment the date and put us right back where we started!

https://github.com/onc-healthit/inferno-program/blob/cda7ed0acf3a2de8ec2799bca08903964374f54f/lib/app/sequence_base.rb#L844

So by adding or subtracting 2 instead of one it actually increments or decrements regardless of a previous increment or decrement.

radamson: @czh-orz is working on a change now that will pull the date from the resource directly in `date_comparator_value` so that it doesn't depend on a previous search parameter calculation

czh-orz: Thanks for the explanation Reece. Yeah, I just updated it so that the comparator search will pull from the original resource instead of from the search params.

radamson: No problem, thank you for explaining it to me originally (and actually making this stuff work)!

PR #140, opened 2020-07-08
Remove warnings for optional metadata in SMART discovery sequence.

arscan:
The SMART discovery sequence current provides warnings when optional fields are not provided in the two metadata endpoints.  These optional fields are unlikely to be implemented by anyone undergoing certification, and it is even less likely they will point to standardized capabilities because there is little to no guidance on how to implement them.  Therefore, do not flag their absence as a warning in the Program Edition.

Note: as an example, our reference implementation does not currently implement them, because it is unclear exactly what is the right way to do so.

**Submitter:**
- [X] Internal ticket for this PR:

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/onc_program/onc_smart_discovery_sequence.rb
+++ b/lib/modules/onc_program/onc_smart_discovery_sequence.rb
@@ -17,32 +17,22 @@ module Inferno
 
         The #{title} Sequence test looks for authorization endpoints and SMART
         capabilities as described by the [SMART App Launch
-        Framework](http://hl7.org/fhir/smart-app-launch/conformance/index.html).
-        The SMART launch framework uses OAuth 2.0 to *authorize* apps, like
-        Inferno, to access certain information on a FHIR server. The
-        authorization service accessed at the endpoint allows users to give
-        these apps permission without sharing their credentials with the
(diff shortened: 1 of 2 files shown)


arscan:
@Jammjammjamm -- do you have thoughts on this one?  I'm cleaning up warnings against our reference server because I think we should strive to demonstrate an API with no warnings.  In this case, it just seems harsh to flag a warning for empty urls with no real guidance as to how to properly implement them.  Certainly we wouldn't want to just put in urls to nowhere (even though that will eliminate the warning).  And I'm not sure if we would want to put in URLs to services that don't follow a standard, but we have no guidance on what standard to use.

Jammjammjamm:
I think those warnings were added in the context of a generic SMART test rather than something program specific. I'm not sure which specific capabilities are generating these warnings, but given our overall approach of not issuing errors/warnings for optional functionality in the program tests, removing them makes sense to me.

SMART also lists some things as recommended rather than optional (not sure if that is relevant here), so there is a stronger argument for including a warning there (although perhaps not in program).

arscan:
Ok, since no objection I'm going to merge this to cut down on warnings where there is limited guidance on what to do with the warning.

PR #141, opened 2020-07-08
Add resource_contains_category method

yunwwang:
Create method resource_contains_categroy in validation.rb to check if a resource has required category.code and categroy.system

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/utils/validation.rb
+++ b/lib/app/utils/validation.rb
@@ -196,7 +196,7 @@ module Inferno
     def self.resource_contains_category(resource, category_code, category_system = nil)
       resource&.category&.any? do |category|
         category.coding&.any? do |coding|
-          coding.code == category_code && (category_system.blank? || category_system == coding.system)
+          coding.code == category_code && (category_system.blank? || coding.system.blank? || category_system == coding.system)
         end
       end
     end
--- a/lib/app/utils/validation.rb
+++ b/lib/app/utils/validation.rb
@@ -162,7 +162,7 @@ module Inferno
       if resource.resourceType == 'Observation'
         return DEFINITIONS[US_CORE_R4_URIS[:smoking_status]] if observation_contains_code(resource, '72166-2')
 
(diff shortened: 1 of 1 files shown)


lib/app/utils/validation.rb
```diff
@@ -192,5 +192,21 @@ def self.guess_r4_profile(resource)
     def self.observation_contains_code(observation_resource, code)
       observation_resource&.code&.coding&.any? { |coding| coding&.code == code }
     end
+
+    def self.resource_contains_category(resource, category_code)
```
arscan, line 196: Would it make more sense to have the signature be `def self.resource_contains_category(resource, category_system, category_code)` so you don't need to have that `if` statement that will need to be updated if there is ever another resource type?

yunwwang: I had that version before this one. Then I think why I force the caller to provide category_system while the function itself can decide. It is a problem when the binding value set has codes from multiple code system. At least for Observation and DiagnosticReport, both binding ValueSets have single code system value.

yunwwang: I fixed it.

lib/app/utils/validation.rb
```diff
@@ -192,5 +192,13 @@ def self.guess_r4_profile(resource)
     def self.observation_contains_code(observation_resource, code)
       observation_resource&.code&.coding&.any? { |coding| coding&.code == code }
     end
+
+    def self.resource_contains_category(resource, category_code, category_system = nil)
+      resource&.category&.any? do |category|
+        category.coding&.any? do |coding|
+          coding.code == category_code && (category_system.blank? || category_system == coding.system)
```
arscan, line 199: If the system leaves the `system` blank but fills in the `code`, is that ok?  This obviously requires the system to be filled in otherwise it won't match.

yunwwang: good catch!

yunwwang: fixed

PR #145, opened 2020-07-21
FI-920 Fix verify check.

arscan:
This update refines our 'Access verify' set of tests.  The intent of these tests are to ensure that patients have control over what scopes are granted to the application.  We accomplish this by having one sequence that requires full scopes, and verifies that the scopes returned by the server have all the required scopes to get all USCDI references.  And another sequence that prompts the tester to choose a subset of scopes (defaulting to Observation, Condition and Patient), and then the tester is supposed to only grant those scopes, and Inferno verifies that is the case for what is returned.

Previously, these tests passed more than they should have (you could allow for extra scopes in certain cases).  This PR much more thoroughly ensures that the logic is correct, and contains a number of self-tests to ensure that they are.

To try it out, on the 'Single Patient App' set of tests, it should pass if you allow all scopes.  It should pass if you deny 'PractitionerRole, Location, RelatedPerson, Medication' on our reference server as well, because implementations could technically NOT implement 'read' for those resources so we provide flexibility there.  But it should fail if you uncheck any others.

Then, try out the Limited Patient App set of tests.  For this, it should pass if you deny everything except patient/launch and the resources you selected (by default Condition, Observation, Patient).  It should also pass if you allow those that you chose as well as Practitioner, PractitionerRole, Location, RelatedPerson, Encounter and Medication, because it is up to the SUT to decide if those other resources are needed and there doesn't ahve to be an explicit user option to allow/deny them.  Also, it should fail if you do not check off `offline_access`, because we are checking to make sure that it is there in this scenario.

**Submitter:**
- [x] Internal ticket for this PR: FI-920

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/templates/access_verify_sequence.rb.erb
+++ b/generator/uscore/templates/access_verify_sequence.rb.erb
@@ -155,18 +155,18 @@ module Inferno
        assert @instance.received_scopes.split(' ').exclude?('offline_access'), 'This test expects the user to deny offline access to demonstrate that refresh tokens require user approval'
        pass "Resources to be denied: #{denied_resources.join(',')}"
        <% else %>
-        # Consider all directly-mapped USCDI resources, as well as Encounter, Practitioner and Organization
-        # because they have US Core Profile references in the other US Core Profiles.  This excludes
-        # PractionerRole, Location and RelatedPerson because they do not have US Core Profile references
-        # and therefore could be 'contained' and do not have a read interaction requirement.
-        all_resources = [
-        <%= non_delayed_sequences.group_by{|sequence| sequence[:resource]}.values.map(&:first).map{|resource| "'#{resource[:resource]}'"}.join(",\n") %>,
-        'Patient',
-        'Provenance',
-        'Encounter',
-        'Practitioner',
-        'Organization'
-        ]
+       # Consider all directly-mapped USCDI resources, as well as Encounter, Practitioner and Organization
+       # because they have US Core Profile references in the other US Core Profiles.  This excludes
+       # PractionerRole, Location and RelatedPerson because they do not have US Core Profile references
+       # and therefore could be 'contained' and do not have a read interaction requirement.
+       all_resources = [
+       <%= non_delayed_sequences.group_by{|sequence| sequence[:resource]}.values.map(&:first).map{|resource| "'#{resource[:resource]}'"}.join(",\n") %>,
+       'Patient',
+       'Provenance',
+       'Encounter',
+       'Practitioner',
+       'Organization'
+       ]
        allowed_resources = all_resources.select {|resource| scope_granting_access(resource, @instance.received_scopes).present?}
        denied_resources = all_resources - allowed_resources
        assert denied_resources.empty?, "This test requires access to all US Core resources with patient information, but the received scope '#{@instance.received_scopes}' does not grant access to the '#{denied_resources.join(', ')}' resource type(s)."
--- a/generator/uscore/templates/access_verify_sequence.rb.erb
+++ b/generator/uscore/templates/access_verify_sequence.rb.erb
@@ -66,8 +66,8 @@ module Inferno
         resource type is accessed by queries through other resource types. These resources types are accessed in the more
         comprehensive Single Patient Query tests.
 
-        However, the authorization system must indicate that access is granted to the Encounter, Practitioner and Location
-        resource types because they are required to support the read interaction. 
+        However, the authorization system must indicate that access is granted to the Encounter, Practitioner and Organization
+        resource types by providing them in the returned scopes because they are required to support the read interaction. 
       )
       <% end %>
       requires :onc_sl_url, :token, :patient_id, :received_scopes<% if access_verify_restriction == 'restricted' %>, :onc_sl_expected_resources<% end %>
--- a/generator/uscore/templates/access_verify_sequence.rb.erb
+++ b/generator/uscore/templates/access_verify_sequence.rb.erb
@@ -108,7 +108,7 @@ module Inferno
 
       def scope_granting_access(resource, scopes)
         scopes.split(' ').find do |scope| 
-          scope.start_with?("patient/#{resource}.", 'patient/*.') && scope.end_with?('.*', '.read')
+          ['patient/*.read', 'patient/*.*', "patient/#{resource}.read", "patient/#{resource}.*"].include? scope
         end
       end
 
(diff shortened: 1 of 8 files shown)


generator/uscore/templates/access_verify_sequence.rb.erb
```diff
@@ -60,6 +65,9 @@ module Inferno
         Organization, Practitioner, PractionerRole, and RelatedPerson.  It also does not test Provenance, as this
         resource type is accessed by queries through other resource types. These resources types are accessed in the more
         comprehensive Single Patient Query tests.
+
+        However, the authorization system must indicate that access is granted to the Encounter, Practitioner and Location
```
Jammjammjamm, line 69: What's the deal with this?

generator/uscore/templates/access_verify_sequence.rb.erb
```diff
@@ -100,7 +108,7 @@ module Inferno
 
       def scope_granting_access(resource, scopes)
         scopes.split(' ').find do |scope| 
-          scope.start_with?("patient/#{resource}", 'patient/*') && scope.end_with?('*', 'read')
+          scope.start_with?("patient/#{resource}.", 'patient/*.') && scope.end_with?('.*', '.read')
```
Jammjammjamm, line 111: This would say `patient/Condition...blah...read` is ok. We should probably be more strict.
```
Regexp.new("patient\\/(\\*|#{resource})\\.(\\*|read)").match? scope
```
Or rather than putting the regex inline, define it in a method somewhere so we can unit test it.

Alternatively, we could just check all four options explicitly:
```
['patient/*.read', 'patient/*.*', "patient/#{resource}.read", "patient/#{resource}.*"].include? scope
```

generator/uscore/templates/access_verify_sequence.rb.erb
```diff
@@ -112,29 +120,55 @@ module Inferno
...
+       received_scope_resources = all_resources.select{|resource| scope_granting_access(resource, @instance.received_scopes).present?}
        unexpected_resources = received_scope_resources - allowed_resources
        assert unexpected_resources.empty?, "This test expected the user to deny access to the following resources that are present in scopes received during token exchange response: #{unexpected_resources.join(', ')}"
+       improperly_denied_resources = allowed_resources.reject { |resource| scope_granting_access(resource, @instance.received_scopes).present? }
+       assert improperly_denied_resources.empty?, "This test expected the user to grant access to the following resources that are not received during token exhange response: #{improperly_denied_resources.join(', ')}"
+       assert @instance.received_scopes.split(' ').exclude?('offline_access'), 'This test expects the user to deny offline access to demonstrate that refresh tokens require user approval'
        pass "Resources to be denied: #{denied_resources.join(',')}"
        <% else %>
+        # Consider all directly-mapped USCDI resources, as well as Encounter, Practitioner and Organization
```
Jammjammjamm, line 158: It looks like lines 158-69 are indented an extra space.

generator/uscore/templates/access_verify_sequence.rb.erb
arscan: First off, I have a typo in here.  It should be Encounter, Practitioner and Organization (not Location).

This is complicated.  I would love more scrutiny / double checking on this.  As we know, Encounter, Organization, Practitioner, PractitionerRole, Location and RelatedPerson are not directly mapped into USCDI.  However, they are needed in order to completely implement those US Core resources that are directly mapped into USCDI.

Encounter, Organization and Practitioner have at least one 'US Core typed reference' in another US Core profile.  Because of that, you need to support their US Core profile.  And because of that, you really need to support the whole resource, and it makes sense for it to 'stand alone with its own identity', and therefore you need to support the read interaction on it (vs just containing it in resources that reference it).  Therefore, you need to be able to provide scopes for those, either through patient/*.read or explicitly on those resources.  For the scenario where we expect full access, we need to see those scopes.

For PractitionerRole, Location and RelatedPerson, since there are no explicit US core typed references to them in other US Core resources, you can just 'contain' them in their referring resource (similar to Medication).  And therefore we do not need to see their scopes because they may or may not decide to implement them in a standalone way, depending on if those concepts stand alone in their system or not.

Now, to be completely transparent on what this test does, we do check the scopes returned for Practitioner, Organization, and Encounter, but we do not actually perform read requests on ANY of those 6 in this section because it is difficult to find all the needed references.  So we just say 'good enough' on the directly mapped USCDI resources for the actual FHIR searches/reads here, because we do a more thorough job on the complete US Core set of tests.

And FINALLY, what even complicates this further, is that on the 'limited scope' version of all this (vs. the version where we expect to see ALL scopes), the SUT has flexibility to not given the user direct control of authorizing these non-USCDI-mapped resources, so we cannot fail them if they don't provide the Practitioner, Organization and Encounter scopes because it is up to the EHR to decide if given the limited choices the user selected that those need to be included as scopes.

generator/uscore/templates/access_verify_sequence.rb.erb
arscan: Good point.  I like the alternate form because it is more explicit to me.  I added tests for the `scope_granting_access` method to expose this issue and updated to your alternate form.

generator/uscore/templates/access_verify_sequence.rb.erb
arscan: Fixed.

generator/uscore/templates/access_verify_sequence.rb.erb
Jammjammjamm: I think this makes sense. There are references to these specific profiles, so you have to implement the required RESTful interactions for those profiles. I don't think that the logic behind this is clear in the test details, though.

Jammjammjamm: So, just to double check, for Encounter, Organization, Practitioner, PractitionerRole, Location, and RelatedPerson, the expected behavior is to pass the SUT if they provide scopes for these even if they aren't listed in the expected resource grants.

arscan: For the 'unrestricted access' sequence, we need to see Encounter, Organization, Practitioner in the scopes that are granted during the token exchange.  We've set the rule in that scenario that we want them to show us "everything", and expect to be granted access to everything that has at least read access.

For the 'restricted access sequence', we do not need to see any of those 6 in the scopes that were granted (received).  But we also don't fail them if we DO see them.  Essentially, the SUT can decide when to grant access to them or not based on which subset of USCDI resources were selected by the user.  The user doesn't have to be given the option to explicitly grant 'Organization', for example (though they could).

This is difficult to communicate succinctly.

Issue #148, opened 2020-07-22
Capability statement validation indicates 'introspect' and 'revoke' security extensions are not defined


darena-mhaque:
**Subject of the issue**
In the USCCAP-02 test, the capability statement validation indicates that the values 'introspect' and 'revoke' are not defined by the extension http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris. However, if we go to the URL where this extension is defined, http://www.hl7.org/fhir/smart-app-launch/StructureDefinition-oauth-uris.html, we can see that the 'introspect' and 'revoke' values are indeed defined.

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 1.0.2

**Steps to reproduce**
Run USCCAP-02 with any fhir server that supports the "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris" extension and has 'introspect' and 'revoke' as url values. You can view our server's metadata by going to: https://fhir-staging.bluebuttonpro.com/inferno-certification/metadata

darena-mhaque:
Here's an image for clarity:
![image](https://user-images.githubusercontent.com/45796503/88232724-5d194100-cc44-11ea-87bd-b1448d0e3770.png)

yunwwang:
Hello darena-mhaque:

We have realized this problem. There is a discussion on zulip https://chat.fhir.org/#narrow/stream/179170-smart/topic/OAuth-URIs.20Extension. We are working with Josh and Grahame to address that.

arscan:
We resolved this in [v1.1](https://github.com/onc-healthit/inferno-program/releases/tag/v1.1.0P), which was released and deployed yesterday.  I confirmed it no longer errors on the capability statement at https://fhir-staging.bluebuttonpro.com/inferno-certification/ (and I confirmed that CapabilityStatement still has the oauth-uri extension in use).

Thanks for the report.  I'm going to close this, please comment here or open a new ticket if you are still seeing issues.

Issue #147, opened 2020-07-22
Capability statement validation indicates group export does not resolve


darena-mhaque:
**Subject of the issue**
In the USCCAP-02 test, the result indicates that the group-export url does not resolve. Here is an image:
![image](https://user-images.githubusercontent.com/45796503/88231531-66091300-cc42-11ea-94b1-4ed61fbb6304.png)

However, the URL this _is_ resolvable. Here is the URL: http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export. You can copy paste this to your browser window or postman and verify that the URL resolves.

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 1.0.2

**Steps to reproduce**
Run USCCAP-02 with any fhir server that defines the "$export" operation. You can view our server's metadata by going to: https://fhir-staging.bluebuttonpro.com/inferno-certification/metadata

arscan:
We resolved this in [v1.1](https://github.com/onc-healthit/inferno-program/releases/tag/v1.1.0P), which was released and deployed yesterday.  I confirmed it no longer errors on the capability statement at https://fhir-staging.bluebuttonpro.com/inferno-certification/ (and I confirmed that it still has the group-export url referenced there).

Thanks for the report.  I'm going to close this, please comment here or open a new ticket if you are still seeing issues.

PR #150, opened 2020-07-23
Add filter for 3 specific validator errors. Add unit test

yunwwang:
This is a fix for Github issues https://github.com/onc-healthit/inferno-program/issues/147 and https://github.com/onc-healthit/inferno-program/issues/148
The root cause is the FHIR validator does not use correct StructureDefinitions defined by SMART IG.
Change in this PR suppresses FHIR validator errors related to these three use cases. The change does not  impact other validation behavior

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/app/utils/hl7_validator.rb
```diff
@@ -33,7 +38,7 @@ def validate(resource, fhir_models_klass, profile_url = nil)
     private
 
     def issues_by_severity(issues, severity)
-      issues.reject { |i| i.code == 'code-invalid' }
+      issues.reject { |i| i.code == 'code-invalid' || DETAILS_FILTER.include?(i.details.text) }
```
arscan, line 41: Would it make sense to have a 'sanitize_issues' method that just does these 2 things?    Then in the method that calls this do something like `sanitized_issues = sanitize_issues(outcome.issue)` followed by a bunch of `xyz = issues_by_severity(sanitized_issues, 'xyz')`.  I just worry a little that if you need to go through this method in order to throw out the ones we want to hide, but if for some reason someone wants all issues (not just fatal, etc), they won't realize that they need to do copy in this `reject` line first.

arscan: Based on further discussion we are going to change this anyhow, so disregard this comment.

arscan:
Based on discussion:
* Change the exact string match to a regular expression
* Change the url resolve filter to work on any url that doesn't resolve
* Downgrade everything to severity of informational in the filter instead of silently hiding them

yunwwang:
fixed.

test/unit/hl7_validator_test.rb
```diff
@@ -9,52 +9,43 @@
...
-  end
 
-  describe 'Validating a bad resource' do
-    it 'Should pass back an error message' do
-      patient = FHIR::Patient.new(gender: 'robot')
+    it 'removes excluded errors' do
+      outcome = load_fixture('hl7_validator_operation_outcome')
 
-      stub_request(:post, "#{@validator_url}/validate")
+      stub_request(:post, @validator_url + '/validate')
```
arscan, line 35: Minor style thing for your awareness, generally Ruby prefers string interpolation over concatenation with `+`.

```suggestion
      stub_request(:post, "#{@validator_url}/validate")
```

test/unit/hl7_validator_test.rb
```diff
@@ -9,52 +9,43 @@
...
+          body: outcome
         )
-        .to_return(status: 200, body: load_fixture('validator_bad_response'))
-      result = @validator.validate(patient, FHIR)
 
-      assert_equal 2, result[:errors].size
-      assert_equal 1, result[:warnings].size
-      assert_equal 1, result[:information].size
+      result = @validator.validate(@resource, FHIR, @profile)
+      assert result[:errors].length == 2
```
arscan, line 46: The benefit of `assert_equal` here over `assert` is that the error message is more descriptive (so it can say `expected 1 but was given 3` instead of `expected false but was given true`).

lib/app/utils/hl7_validator.rb
```diff
@@ -19,23 +24,32 @@ def validate(resource, fhir_models_klass, profile_url = nil)
...
+      warnings = []
+      information = []
+
+      issues.each do |iss|
+        if iss.severity == 'information' || iss.code == 'code-invalid' || DETAILS_FILTER.any? { |filter| filter.match?(iss&.details&.text) }
+          information << "#{issue_location(iss)}: #{iss&.details&.text}"
+        elsif iss.severity == 'warning'
+          warnings << "#{issue_location(iss)}: #{iss&.details&.text}"
+        else
+          errors << "#{issue_location(iss)}: #{iss&.details&.text}"
```
jason-crowley, line 44: Maybe we should create `"#{issue_location(iss)}: #{iss&.details&.text}"` so we don't have to update it in multiple places if we change it.

radamson: I think creating a method like `issue_message` would be a good idea.

lib/app/utils/hl7_validator.rb
```diff
@@ -3,6 +3,11 @@
 module Inferno
   # A validator that calls out to the HL7 validator API
   class HL7Validator
+    DETAILS_FILTER = [
+      %r{Sub-extension url 'introspect' is not defined by the Extension http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris},
+      %r{Sub-extension url 'revoke' is not defined by the Extension http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris},
+      /URL value .* does not resolve/
```
jason-crowley, line 9: I pasted `http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris` into my browser and got a 404, is this expected? I don't know too much about these filters, but is it worth combining the first two Regexs by replacing them with `%r{Sub-extension url '(introspect|revoke)' is not defined by the Extension http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris}`? Are there other possible values that could go in those quotes?

lib/app/utils/hl7_validator.rb
jason-crowley, line 44: Maybe we should create `"#{issue_location(iss)}: #{iss&.details&.text}"` once so we don't have to update it in multiple places if we change it?

lib/app/utils/hl7_validator.rb
radamson, line 9: Why did you use `/.../` here and `%r{}` above?  I think you're trying to do a regex in both and being consistent would be less confusing.

lib/app/utils/hl7_validator.rb
```diff
@@ -3,6 +3,11 @@
 module Inferno
   # A validator that calls out to the HL7 validator API
   class HL7Validator
+    DETAILS_FILTER = [
```
radamson, line 6: nit: I think `ISSUES_FILTER` better expresses the intention and is consistent with the naming used through the rest of the file (e.g. `issues_by_severity`)
```suggestion
    ISSUES_FILTER = [
```

lib/app/utils/hl7_validator.rb
jason-crowley: I think it's because the URL contains `/` characters

jason-crowley: I'm guessing that the URL doesn't lead anywhere because it's a canonical URL, is that correct?

jason-crowley: Some other relevant changes to the regexes might be to escape the `.` characters with `\.` and to begin and end the regexes with `^` and `$`. I'm not saying these changes have to be made, I'm just making sure we're considering them

lib/app/utils/hl7_validator.rb
yunwwang: My concern is that this is for filtering detail text. It is used at condition expression `|filter| filter.match?(iss.details.text)`
Filters other issue elements, such as expression, or code, have different filter expression.

yunwwang: I changed the name to `ISSUE_DETAILS_FILTER`

lib/app/utils/hl7_validator.rb
```diff
@@ -19,23 +24,32 @@ def validate(resource, fhir_models_klass, profile_url = nil)
...
+        else
+          errors << "#{issue_location(iss)}: #{iss&.details&.text}"
+        end
+      end
+
+      {
+        errors: errors,
+        warnings: warnings,
+        information: information
+      }
```
radamson, line 52: We can cut this method down in length and make it more readable (IMO) by using `#each_with_object`

```suggestion
      issues_hash = { errors: [], warnings: [], information: [] }
      issue_map = { fatal: :errors, error: :errors, warning: :warnings, information: :information }

      issues.each_with_object(issues_hash) do |iss, hsh|
        if iss.code == 'code-invalid'
          issues_hash[:information] << "#{issue_location(iss)}: #{iss&.details&.text}"
        elsif DETAILS_FILTER.any? { |filter| filter.match? iss&.details&.text}
          issues_hash[:information] << "#{issue_location(iss)}: #{iss&.details&.text}"
        elsif issue_map.keys.include? iss.severity.to_sym
        issues_hash[issue_map[iss.severity.to_sym]] << "#{issue_location(iss)}: #{iss&.details&.text}"
        end
      end
```

radamson: I discussed with @arscan and might see if we can do this in a functional way similar to how it was done before...

Might be worth saving a more major refactor for a future PR to get this in quick.

lib/app/utils/hl7_validator.rb
radamson: `http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris` not resolving is ok because the validator can already has the definition or can get it through another means like the package server.

It is related to this [chat.fhir conversation](https://chat.fhir.org/#narrow/stream/179170-smart/topic/OAuth-URIs.20Extension) where the definition is different in SMART than in base FHIR.  we've seen issues where the validator picks up the one from the base FHIR spec which isn't correct.

radamson: We don't want to escape `.` here because `.*` is saying "zero or more of any single character"

arscan: Well you want to escape some of them, but not `.*` (probably what Jason meant).

jason-crowley: Oops, yeah I didn't mean `.*`, but the ones in the url

yunwwang: rubocop prefers that. It won't let me change them consistently.

yunwwang: %r{} escapes all special character. /a\.b/ is the same as %r{a.b}

arscan: yeah looks like rubycop requires `%r{}` if there is a `\`, but requires `\` if there isn't.  I disagree with that!  Just always allow `%r{}`.

lib/app/utils/hl7_validator.rb
yunwwang: fixed

lib/app/utils/hl7_validator.rb
arscan: The problem is that this will match more than just that string, so to be thorough you should escape `.`, and prevent it from having any characters before or after the string, etc.  I think it is good to be thorough, but if you know if matches at least the string you want I also think that there is low likelyhood you will match something you don't want to match in this case.

radamson: I don't think that's correct @yunwwang, I just tested and `%r{}` does not escape periods (it looks like it only escapes `/`.  @jason-crowley is right those need to be escaped.

I went into the `hl7_validator_operation_outcome.json` and changed the `.` to a random character and the tests still passed even though they should have failed (i.e. it is considering the `.` as any single character)

yunwwang: fixed

config.yml
```diff
@@ -39,7 +39,7 @@ badge_text: PROGRAM EDITION
 
 # Resource validator options: must be one of "internal" or "external". external_resource_validator_url is only used if resource_validator is set to external.
 resource_validator: external
-external_resource_validator_url: http://validator_service:4567
+external_resource_validator_url: http://localhost:8888
```
arscan, line 42: Looks like this was committed by mistake.

```suggestion
external_resource_validator_url: http://validator_service:4567
```

yunwwang: Ahhh. How could I get this file checked in again!

PR #158, opened 2020-08-10
Update program procedure to match tp v1.1

arscan:
Updates program matrix to match changes introduced in the [TP v1.1](https://www.healthit.gov/test-method/standardized-api-patient-and-population-services#ccg)

### DIFF
--- a/lib/modules/onc_program_procedure.yml
+++ b/lib/modules/onc_program_procedure.yml
@@ -309,20 +309,14 @@ procedure:
           * "Device";
           * "DiagnosticReport";
           * "DocumentReference";
-          * "Encounter";
           * "Goal";
           * "Immunization";
-          * "Location";
           * "Medication" (if supported);
           * "MedicationRequest";
           * "Observation";
-          * "Organization";
           * "Patient";
-          * “Practitioner";
-          * “PractitionerRole";
-          * “Procedure";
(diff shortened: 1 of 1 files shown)


lib/modules/onc_program_procedure.yml
```diff
@@ -340,20 +334,14 @@ procedure:
           * "Device";
           * "DiagnosticReport";
           * "DocumentReference";
-          * "Encounter";
           * "Goal";
           * "Immunization";
-          * "Location";
```
okeefm, line 346: The `Location` element seems to be present in the TLV list (though it's not in the SUT list)

arscan: Is this an issue in this file?  I'm having trouble finding where Location is on the TLV side, but I might just be missing it.  I do know that ONC's test procedure is wrong, in that Location is on the TLV side on this, but I think I fixed it?

arscan: I'm not seeing this issue, so I'm going to go ahead and resolve this.  The fact that the TP itself has this problem in it means that we either fix it and be inconsistent with the TP, or leave it to be consistent with the TP (but wrong).  So either way it isn't a great situation.  I've reported it to ONC so hopefully in the next version it will be addressed.

PR #165, opened 2020-08-26
Fi 971 must support reference types

czh-orz:
This PR makes it so that we now check to see all us core reference types in reference elements that are "must support".  This means that if reference element can be either reference to a us core patient or us core practitioner, then we expect to see both reference types from the resources returned in the search. 

To note, medicationrequest.reported can either be a boolean or a reference. Before, we were only checking if either one was present. Now, it is expecting to see a reportedReference in there. 

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-971

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -86,6 +86,7 @@ module Inferno
           must_supports: {
             extensions: [],
             slices: [],
+            references: [],
             elements: []
           },
           mandatory_elements: [],
@@ -283,7 +284,6 @@ module Inferno
                 id: element['id'],
                 url: element['type'].first['profile'].first
               }
-            next
           elsif element['sliceName'].present?
             array_el = profile_elements.find { |el| el['id'] == element['path'] }
             discriminators = array_el['slicing']['discriminator']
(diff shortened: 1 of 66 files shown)


generator/uscore/metadata_extractor.rb
```diff
@@ -330,23 +330,34 @@ def add_must_support_elements(profile_definition, sequence)
...
-          elsif element['fixedCode'].present?
-            must_support_element[:fixed_value] = element['fixedCode']
-          elsif element['patternIdentifier'].present?
-            must_support_element[:fixed_value] = element['patternIdentifier']['system']
-            must_support_element[:path] += '.system'
+          elsif element['type'].any? { |type| type['code'] == 'Reference' }
+            reference_type = element['type'].find { |type| type['code'] == 'Reference' }
+            profiles = reference_type['targetProfile']
+            resource_types = profiles
+              .select { |profile| resources_by_type['StructureDefinition'].find { |stdef| stdef['url'] == profile } }
```
Jammjammjamm, line 337: This selects only the US Core references because those are the only profiles we have in `resources_by_type`?

generator/uscore/uscore_generator.rb
```diff
@@ -583,40 +587,31 @@ def create_must_support_test(sequence)
...
 
-        sequence[:must_supports][:elements].each do |element|
-          test[:description] += %(
-            * #{element[:path]})
-          # class is mapped to local_class in fhir_models. Update this after it
-          # has been added to the description so that the description contains
-          # the original path
-          element[:path] = element[:path].gsub(/(?<!\w)class(?!\w)/, 'local_class')
-        end
+            #{must_support_list.join("\n            ")}
```
Jammjammjamm, line 591:
```suggestion
            #{must_support_list.sort.join("\n            ")}
```
I just noticed that these weren't in alphabetical order, which seems odd.

generator/uscore/uscore_generator.rb
```diff
@@ -583,40 +587,31 @@ def create_must_support_test(sequence)
...
 
-        must_support_extensions = sequence[:must_supports][:extensions]
-        must_support_extensions.each do |extension|
-          test[:description] += %(
-            * #{extension[:id]})
-        end
 
+            For elements of type 'reference' with one or more target profiles from US Core, this test will ensure that at least one of each resource type
```
Jammjammjamm, line 594: Should this only be included for the tests which actually have to perform this checking?

generator/uscore/metadata_extractor.rb
czh-orz: Right, all the structure definitions in the us core resource folder are us core profiles. I also can't filter by url because of the vital sign profiles which are base fhir.

arscan:
Does this work on the bulk tests too?

arscan:
> Does this work on the bulk tests too?

@czh-orz verified they do and I confirmed the results are consistent between the bulk & us core API tests.

generator/uscore/uscore_generator.rb
czh-orz: 👍

generator/uscore/uscore_generator.rb
czh-orz: 👍

lib/modules/uscore_v3.1.0/profile_definitions/us_core_encounter_definitions.rb
```diff
@@ -6,6 +6,24 @@ class USCore310EncounterSequenceDefinitions
...
+          },
+          {
+            path: 'participant.individual',
+            resource_types: [
+              'Practitioner'
+            ]
+          },
+          {
+            path: 'location.location',
+            resource_types: []
```
radamson, line 24: Should this be `Location`?

arscan: No, the path is actually `location.location`, because the `location` element has a `location` field that can store the reference, as well as other fields.  See http://hl7.org/fhir/us/core/STU3.1/StructureDefinition-us-core-encounter.html (the Full View is more useful to see why there is that odd looking `location.location`.

PR #170, opened 2020-08-31
FI-970: Don't count extensions in must support elements

Jammjammjamm:
This branch updates the must support element checking to ignore extensions.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -642,6 +642,8 @@ module Inferno
             missing_must_support_elements = must_supports[:elements].reject do |element|
               #{resource_array}&.any? do |resource|
                 value_found = resolve_element_from_path(resource, element[:path]) { |value| element[:fixed_value].blank? || value == element[:fixed_value] }
+                value_found = value_found.to_hash.reject { |key, _| key == 'extension' } if value_found.respond_to? :to_hash
+
                 value_found.present?
               end
             end
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -641,8 +641,10 @@ module Inferno
           test[:test_code] += %(
             missing_must_support_elements = must_supports[:elements].reject do |element|
               #{resource_array}&.any? do |resource|
-                value_found = resolve_element_from_path(resource, element[:path]) { |value| element[:fixed_value].blank? || value == element[:fixed_value] }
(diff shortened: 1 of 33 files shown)


arscan:
@czh-orz I believe you took an action to review this to see if there was a way to update the block so that we don't just match against the first, but look for anything that isn't only data-absent-reason.

generator/uscore/uscore_generator.rb
```diff
@@ -642,6 +642,8 @@ def create_must_support_test(sequence)
             missing_must_support_elements = must_supports[:elements].reject do |element|
               #{resource_array}&.any? do |resource|
                 value_found = resolve_element_from_path(resource, element[:path]) { |value| element[:fixed_value].blank? || value == element[:fixed_value] }
```
czh-orz, line 644:
```suggestion
                value_found = resolve_element_from_path(resource, element[:path]) do |value|
                  value_without_extensions = value.to_hash.reject { |key, _| key == 'extension' } if value.respond_to? :to_hash
                  value_without_extensions.present? && (element[:fixed_value].blank? || value == element[:fixed_value])
                end
```

czh-orz: `resolve_element_from_path` will only return elements that resolve to true when passed through the code block. So if the element is in an array, then this will look through the array for elements that are more than extensions.

arscan: Moving this here would make it so that the new line 645 `value_found = value_found.to_hash.reject { |key, _| key == 'extension' } if value_found.respond_to? :to_hash` can be removed, right?  Because essentially we have filtered within this block, so we don't need to do it again after.

arscan:
Does this also address bulk data must support checking?

See https://github.com/onc-healthit/inferno-program/blob/79eecdd1377933e7b89741c2850e979826675ddf/lib/modules/onc_program/bulk_data_group_validation_sequence.rb#L206

Also, I think we recently merged changes into the bulk data side of things, so a rebase before addressing this is probably a good idea.

generator/uscore/uscore_generator.rb
Jammjammjamm: I've addressed this in my most recent push. I had to manually add the changes since this code isn't quite right, and as you say, it requires changing multiple lines.

Jammjammjamm:
It does now.

PR #178, opened 2020-09-09
FI-996: Check resolved reference types

Jammjammjamm:
This branch checks that the correct resource type is returned when resolving a reference.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -30,6 +30,8 @@ module Inferno
       include SearchValidationUtil
       include Inferno::WebDriver
 
+      class InvalidReferenceResource < StandardError; end
+
       @@test_index = 0
 
       @@group = {}
@@ -708,10 +710,17 @@ module Inferno
                 next
               end
             end
-            value.read
+            reference = value.reference
+            reference_type = value.resource_type
+            resource = value.read
+
+            raise InvalidReferenceResource if resource.resourceType != reference_type
+
             resolved_references.add(value.reference)
           rescue ClientException => e
             problems << "#{path} did not resolve: #{e}"
+          rescue InvalidReferenceResource
+            problems << "Expected #{reference} to refer to a #{reference_type} resource, but found a #{resource.resourceType} resource."
           end
         end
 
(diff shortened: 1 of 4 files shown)


lib/app/sequence_base.rb
```diff
@@ -708,10 +710,17 @@ def validate_reference_resolutions(resource, resolved_references = Set.new, max_
...
+            reference = value.reference
+            reference_type = value.resource_type
+            resource = value.read
+
+            raise InvalidReferenceResource if resource.resourceType != reference_type
+
             resolved_references.add(value.reference)
           rescue ClientException => e
             problems << "#{path} did not resolve: #{e}"
+          rescue InvalidReferenceResource
```
arscan, line 722: This is an update to the 'platform'.  Is it generic enough to put this here vs. in a test?

Jammjammjamm: Yeah, I think verifying that a reference is the correct type is pretty generic functionality.

lib/app/sequence_base.rb
```diff
@@ -708,10 +710,17 @@ def validate_reference_resolutions(resource, resolved_references = Set.new, max_
                 next
               end
             end
-            value.read
+            reference = value.reference
+            reference_type = value.resource_type
+            resource = value.read
+
+            raise InvalidReferenceResource if resource.resourceType != reference_type
+
```
yunwwang, line 718: Is there a unit test for this newly introduced logic?

Jammjammjamm: I didn't add any unit tests for this, but I did have to update some existing unit tests which were failing because they were returning incorrect resource types.

yunwwang: Is it convenient to add some unit tests for this section?

Jammjammjamm: Added.

lib/app/sequence_base.rb
arscan: Ok, if you think so I'm good with it.  I just worry because it isn't clear to me when you want to specialize the Exception and when you don't.  If we encourage this pattern then this list of rescues will get mighty long.  Should we have an 'InvalidResource' exception (that seems even more generic than this)?  An 'InvalidFhirUser' exception for SMART?

arscan: Just looked a little closer and now realize this isn't the main rescue section of sequence_base but is isolated to this section.  That makes more sense!

arscan (approved):

PR #210, opened 2020-12-01
Fi 1066-either carrieraidc carrierhrf must support

czh-orz:
This PR makes it so that either carrierAIDC or carrierHRF is found in the must support test. I take out both elements from the must support metadata and add a special check for udicarrier.carrierAIDC and udicarrier.carrierHRF. 

This PR also includes a change to head_occipital_frontal_circumference_percentile_sequence which was made by the generator. I'm guessing it was just not included in another PR.

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-1066

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -559,6 +559,11 @@ module Inferno
           (path.start_with? 'component.value') && (!path.include? '[x]') && (path != 'component.valueQuantity')
         end
 
+        # remove carrierAIDC and carrierHRF from implantable device must supports
+        implantable_device_sequence = metadata[:sequences].find { |sequence| sequence[:profile] == 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-implantable-device'}
+          implantable_device_sequence[:must_supports][:elements].delete_if do |element|
+            ['udiCarrier.carrierAIDC', 'udiCarrier.carrierHRF'].include? element[:path]
+          end
         metadata
       end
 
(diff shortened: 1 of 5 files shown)


generator/uscore/metadata_extractor.rb
```diff
@@ -559,6 +559,11 @@ def add_special_cases(metadata)
           (path.start_with? 'component.value') && (!path.include? '[x]') && (path != 'component.valueQuantity')
         end
 
+        # remove carrierAIDC and carrierHRF from implantable device must supports
+        implantable_device_sequence = metadata[:sequences].find { |sequence| sequence[:profile] == 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-implantable-device' }
```
yunwwang, line 563: I think we should have a general solution, like a predefined list of "must-supported element group" which fall into the same pattern, ie all the elements in such group are marked as must-supported but only one of element is required to be supported.
Soon, there will be more of similar groups.

arscan: I think we should keep this simple solution until we need the more complicated one.  Right now this is the only either-or must support.  When v3.2 lands, I think we can address the many either-or must supports.  Part of my motivation here is that I would like a stable solution that addresses today's program certification problem today (which this covers), and I'm concerned that making a stable more general solution will take a little while.

arscan (approved):
I tried it out by removing the carrierHRF in a test patient, which 'skips' in the current version but 'passes' in this version.  Thank you for addressing the textual narrative part (support x or y) too.

I think @yunwwang has a good point, but I think we should merge this now because it meet's today's need and we can address the more general case in the next release.

yunwwang (approved):

PR #212, opened 2020-12-02
FI-1083 Fix invalid invariants in FHIR v4.0.1 and US Core v3.1.1

arscan:
There is an invariant in FHIR and US Core vital signs that has been acknowledged to be invalid.  Hopefully this will be addressed in future versions of US Core, but in the meantime we will 'mute' the error from the validator.

```
vs-1: if Observation.effective[x] is dateTime and has a value then that value shall be precise to the day [($this as dateTime).toString().length() >= 8]

uscore-1: if Observation.effective[x] is dateTime and has a value then that value shall be precise to the day [($this as dateTime).toString().length() >= 8]
```

Note: I recognize that we could probably just key off of the invariant key (uscore-1, vs-1), but I am skeptical that these are globally unique forever, so instead of just keying off them, I do it plus the first N characters of the description (because rubocop gets mad if the line lengths are too long).  I see no downside to this, and in fact it is useful to see in the code a little bit of information about which invariant these are referring to).

**Submitter:**
- [x] Internal ticket for this PR: FI-1083

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/app/utils/hl7_validator.rb
+++ b/lib/app/utils/hl7_validator.rb
@@ -7,8 +7,8 @@ module Inferno
       %r{^Sub-extension url 'introspect' is not defined by the Extension http://fhir-registry\.smarthealthit\.org/StructureDefinition/oauth-uris$},
       %r{^Sub-extension url 'revoke' is not defined by the Extension http://fhir-registry\.smarthealthit\.org/StructureDefinition/oauth-uris$},
       /^URL value .* does not resolve$/,
-      /^vs-1: if Observation.effective\[x\] is dateTime and has a value then that value shall be precise to the day/, # Invalid invariant in FHIR v4.0.1
-      /^uscore-1: if Observation.effective\[x\] is dateTime and has a value then that value shall be precise to the day/ # Invalid invariant in US Core v3.1.1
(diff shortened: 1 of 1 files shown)


lib/app/utils/hl7_validator.rb
```diff
@@ -6,7 +6,9 @@ class HL7Validator
     ISSUE_DETAILS_FILTER = [
       %r{^Sub-extension url 'introspect' is not defined by the Extension http://fhir-registry\.smarthealthit\.org/StructureDefinition/oauth-uris$},
       %r{^Sub-extension url 'revoke' is not defined by the Extension http://fhir-registry\.smarthealthit\.org/StructureDefinition/oauth-uris$},
-      /^URL value .* does not resolve$/
+      /^URL value .* does not resolve$/,
+      /^vs-1: if Observation.effective\[x\] is dateTime and has a value then that value shall be precise to the day/, # Invalid invariant in FHIR v4.0.1
+      /^uscore-1: if Observation.effective\[x\] is dateTime and has a value then that value shall be precise to the day/ # Invalid invariant in US Core v3.1.1
```
Jammjammjamm, line 11: These periods should technically be escaped like in the filters above, but I don't think it really matters as it's difficult to imagine getting false positives from not escaping them.

Jammjammjamm (approved):

PR #227, opened 2021-02-04
Add Content-Type check for metadata response

yunwwang:
This PR add checking Content-Type value for the metadata response. The Content-Type shall be 'application/fhir+json' according to http://hl7.org/fhir/http.html#mime-type

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/onc_program/bulk_data_export_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_export_sequence.rb
@@ -39,6 +39,7 @@ module Inferno
         headers = { accept: 'application/fhir+json' }
         reply = LoggedRestClient.get(url, headers)
         assert_response_ok(reply)
+        assert_response_content_type(reply, 'application/fhir+json')
 
         conformance = versioned_resource_class.from_contents(reply.body)
         assert conformance.present?, 'Cannot read server CapabilityStatement.'
--- a/lib/modules/onc_program/bulk_data_export_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_export_sequence.rb
@@ -28,21 +28,10 @@ module Inferno
       end
 
       def check_capability_statement
-        if @instance.bulk_url.present?
-          url = @instance.bulk_url
-          url = url.chop if url.end_with?('/')
-        else
-          url = ''
-        end
-
-        url += '/metadata'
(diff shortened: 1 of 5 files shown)


lib/modules/onc_program/bulk_data_export_sequence.rb
```diff
@@ -39,6 +39,7 @@ def check_capability_statement
         headers = { accept: 'application/fhir+json' }
         reply = LoggedRestClient.get(url, headers)
```
arscan, line 40: We should use the fhir_client here and the built in capabilities functionality at this point, I think.

lib/modules/us_core_guidance/test/us_core_r4_capability_statement_sequence_test.rb
```diff
@@ -25,6 +25,20 @@
 
       assert test_result.fail?, "Test result should be fail but is #{test_result.result}"
     end
+
+    it 'fail if Content-Type is not application/fhir+json' do
```
arscan, line 29: A minor thing, but I think the convention is to be "fails if...", so the line reads like "it fails if".  I don't feel strongly about it though.  But If you change it, change all the others too.

arscan:
Could you update the description of this specific test to state the capabilities interaction conforms to the requirements in the base FHIR specification, with a link maybe?

https://github.com/onc-healthit/inferno-program/blob/2e5f51ccf322f8487c910352bbc073127c15062a/lib/modules/onc_program/bulk_data_export_sequence.rb#L159

If someone fails and objects, I want the description of the test be where they can look to find out why they failed.  right now the description just says that it must state support for the bulk export operation, which is not quite the same thing.

arscan:
Is this test consistent with the capabilities test for single patient / smart discovery?  Please make sure that this is completely consistent everywhere we are performing the capabilities interaction on the server & update those other spots if needed.  I thought we needed to update those other spots too, but I could be wrong.

lib/modules/onc_program/bulk_data_export_sequence.rb
yunwwang: fixed

yunwwang:
Will continue working on this

PR #228, opened 2021-02-04
Update docker compose

arscan:
If you are using `docker-compose run` to execute inferno from the command line (to run a script, for example), the `depends_on` is very important but wasn't included.

I also renamed the container names from `ruby_server` to `inferno` and `nginx_server` to `nginx` because they are better descriptions of the containers.  Docker may nag you to clean up orphaned containers due to this name change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
   - docker-compose build
 script:
   - bundle exec rake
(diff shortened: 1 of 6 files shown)


okeefm (approved):
Seems like a reasonable change to me.

PR #234, opened 2021-02-12
save only non-operation outcome results

czh-orz:
This PR makes it so that sequences don't save OperationOutcome results after performing a search. If the search was successful, and an operationoutcome was returned the bundle, the sequence will continue as if it didn't see an OperationOutcome. 

This change won't affect the situation where a search fails and they return an operation outcome. 

**Submitter:**
- [X] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/FI-1149

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
+++ b/lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
@@ -265,10 +265,6 @@ module Inferno
           reply = perform_search_with_status(reply, search_params) if reply.code == 400
 
           validate_search_reply(versioned_resource_class('Observation'), reply, search_params)
-          resource_returned = fetch_all_bundled_resources(reply, check_for_data_absent_reasons)
-          assert(resource_returned.all? { |resource| ['Observation', 'OperationOutcome'].include? resource.resourceType },
-                 'All resources returned must be of the type Observation or OperationOutcome')
-          resource_returned.reject! { |resource| resource.resourceType == 'OperationOutcome' }
 
           ['gt', 'ge', 'lt', 'le'].each do |comparator|
             comparator_val = date_comparator_value(comparator, resolve_element_from_path(@observation_ary[patient], 'effective') { |el| get_value_for_search_param(el).present? })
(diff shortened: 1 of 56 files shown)


arscan:
We need to explicitly fail if it is something other than the expected resource type prior to validation or must support tabulation.  Because if they do incorrectly still send the wrong resource type, our code ungracefully errors (purple error with cryptic ruby error) instead of stating that it received an unexpected resource type.

Also, I don't think this is right spot to strip this out -- I'd put it closer to the actual test (this seems more like a helper function to page through all resources, I wouldn't expect a function named this to silently remove OperationOutcomes).

lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
```diff
@@ -259,6 +265,10 @@ def patient_ids
           reply = perform_search_with_status(reply, search_params) if reply.code == 400
 
           validate_search_reply(versioned_resource_class('Observation'), reply, search_params)
+          resource_returned = fetch_all_bundled_resources(reply, check_for_data_absent_reasons)
```
arscan, line 268: Is there any way to do this without paging through all the results again?  Our tests are already pretty slow, and this adds more requests.  Can we just move this assert into `validate_search_results`?

lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
```diff
@@ -259,6 +265,10 @@ def patient_ids
...
+          assert(resource_returned.all? { |resource| ['Observation', 'OperationOutcome'].include? resource.resourceType },
+                 'All resources returned must be of the type Observation or OperationOutcome')
+          resource_returned.reject! { |resource| resource.resourceType == 'OperationOutcome' }
```
arscan, line 271: What does `resource_returned` do?  It doesn't appear to be referenced again, so why clear out OperationOutcomes from it?

arscan: Ah, I think we have it here because it is needed for MedicationRequest.  But I think we could have the `if` statement in the generator be a little more inclusive so that this line doesn't show up in the generated code for no reason for non-MedicaitonRequests.

arscan: hmm, in looking at the generator code, I can see why it is a little awkward to have a multi-line if block here.

czh-orz: Yeah, I changed it to only add this for medication inclusion. I also just moved the all resources are the right resource check into validate_search_reply

Jammjammjamm (approved):

PR #238, opened 2021-02-17
FI-1150 skip contained reference from save_delayed_reference

yunwwang:
# Summary
Add logic skipping contained reference from treating as delayed references

## New behavior
reference to contained resource are skipped

## Code changes
save_delayed_sequence_references() in sequence_base.rb

## Testing guidance
Add a MustSpported reference to a contained resource. Example provided at lib/module/uscore_v3.1.1/test/fixutres/us_core_careteam.json

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -39,7 +39,7 @@ resource_validator: external
 
 # The url where the external validator can be reached. Ignored when using the
 # internal validator.
-external_resource_validator_url: http://validator_service:4567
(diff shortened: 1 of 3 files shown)


Jammjammjamm (approved):

arscan (approved):
I confirmed that our unit tests failed with that example resource, and your fix prevents that from happening again.  The surface area here is small enough that I don't think there can be any unintended side effects.

PR #239, opened 2021-02-17
FI-1157 Allow more flexibility in which endpoints expose which smart capabiliites

arscan:
# Summary

If a system implements patient-centric and practitioner-centric APIs using different FHIR endpoints, then they should not need to state support for all smart capabilities at every endpoint.

## New behavior

This now only ensures that the capabilities relevant to the current target use case (patient standalone vs practitioner ehr) is tested.

## Code changes

The only tricky part here is how to override what used to be a CONSTANT.  Please verify that this approach makes sense.

## Testing guidance

Review updated tests.  They should cover all necessary cases.

### DIFF
--- a/lib/modules/onc_program/onc_smart_discovery_sequence.rb
+++ b/lib/modules/onc_program/onc_smart_discovery_sequence.rb
@@ -47,7 +47,27 @@ module Inferno
         # run after the oauth endpoints are saved
       end
 
-      REQUIRED_SMART_CAPABILITIES = [
+      # Only require EHR launch related capabilities
+      # A separate sequence handles standalone launch
+
+      def self.required_smart_capabilities
+        [
+          'launch-ehr',
+          'client-confidential-symmetric',
+          'sso-openid-connect',
+          'context-banner',
+          'context-style',
+          'context-ehr-patient',
+          'permission-offline',
+          'permission-user'
+        ]
+      end
+
+      def required_smart_capabilities
+        self.class.required_smart_capabilities
+      end
+
+      SMART_CAPABILITIES = [
         'launch-ehr',
         'launch-standalone',
         'client-public',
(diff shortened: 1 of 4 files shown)


lib/modules/onc_program/onc_smart_discovery_sequence.rb
```diff
@@ -267,7 +287,7 @@ def after_save_oauth_endpoints(oauth_token_endpoint, oauth_authorize_endpoint)
           description %(
             A SMART on FHIR server SHALL convey its capabilities to app
             developers by listing a set of the capabilities. The following
-            capabilities are required: #{REQUIRED_SMART_CAPABILITIES.join(', ')}
+            capabilities are required: #{OncSMARTDiscoverySequence.required_smart_capabilities.join(', ')}
```
Jammjammjamm, line 290: This shows the same list of required capabilities for both standalone and ehr launches:

Standalone:
Screen Shot 2021-02-18 at 8 34 38 AM

EHR:
Screen Shot 2021-02-18 at 8 35 16 AM

arscan: Ah, good catch.  Unfortunately I'm not sure if there is any great way of fixing this without changing the code sharing strategy here, since this is executed within the context of a Test and doesn't know which sequence it is a part of.  I think the best fix here is for us to just describe both here.  e.g. "For patient launches, we are checking these capabilities ..., for ehr practitioner launches we are testing these other ones ...".  Thougths?

Jammjammjamm: I think that is an appropriate solution. Otherwise it would be a lot of work just to change the description.

arscan: It turns out this is quite a pain to list the capabilities, because the 'parent' sequence doesn't have access to the 'child' sequence's required capabilities for the description, and requiring us to load that file caused some wierd behavior.  So I just left the language somewhat vague as to exactly which are required in each particular scenario in the description itself.  I think it should be self-evident based on errors, if they come up.  Does that sound reasonable?

Jammjammjamm: Yeah, sure.

Jammjammjamm (approved):

PR #241, opened 2021-02-19
Prefill bearer token on reference server for ease of use.

arscan:
# Summary
We recently upgraded our reference server to allow users to preconfigure a static bearer token that always work.  Update the config to prefill that for the reference server for ease of use purposes fro the 'inferno.healthit.gov' reference server preset.  Note that this will technically not work until the new reference server is deployed, but since it isn't prefilled now anyhow, this doesn't really hurt things.

See https://github.com/onc-healthit/inferno.healthit.gov/pull/71

## New behavior

Bearer token for single patient api tests will be prefilled for the 'inferno.healthit.gov' reference server preset.


okeefm (approved):
Looks good. Tested, the token shows up.

PR #242, opened 2021-02-19
unescape the \\, in search value

yunwwang:
# Summary
when creating search_parameter, comma(,) character is replaced with "\\," to avoid conflict with FHIR multiple OR search. For example: "Dole,John" is reformed search parameter: name=Dole\\,John in order to distinguish from name=Dole,John which is searching name starts with "Dole" or "John"

While compare the result with search parameter, the "\\," in search parameter need to changed back to "," only.

## New behavior
Unescape "\\," in search value to ","

## Code changes
validate_reply_entries in sequence_base.rb
new unit test
updated testing artifact with name text

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -537,7 +537,8 @@ module Inferno
           assert resource_validation_errors[:errors].empty?, "Invalid #{resource.resourceType}: #{resource_validation_errors[:errors].join("\n* ")}"
 
           search_params.each do |key, value|
-            validate_resource_item(resource, key.to_s, value)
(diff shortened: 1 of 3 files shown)


arscan (approved):
I confirmed that inserting a `\` into an existing patient name on our reference server caused the incorrect failure, and that this PR eliminates that incorrect failure.
Screen Shot 2021-02-22 at 1 47 02 PM
I also double-checked that this fix mirrors the encoding we do here: https://github.com/onc-healthit/inferno-program/blob/main/lib/app/sequence_base.rb#L851

PR #243, opened 2021-02-23
Delay failure to enable more tests to run.

arscan:
An issue that I came across is that in this implementation, sending an incorrect resource type causes the test to fail prior to saving any of the resources.  This is not a great experience for the user, since none of the later tests in the sequence will be able to run.  We try to avoid that.

Does this fix that problem @czh-orz?


czh-orz (approved):
Nice, I think it looks good and gives more info on what the extra resource types are.

arscan:
Ok, I'm merging into your PR and we can look at that one more time as a whole before approving that.

PR #246, opened 2021-03-04
FI 1175 validator updates

okeefm:
# Summary

* Update validator version to 1.2.0 in compose, and remove filter for URL Does Not Resolve errors
* Add an expected validator version to config.yml, and display a (hopefully) helpful error if it doesn't match the fhir-validator-service version

## New behavior

Updated to the latest fhir-validator-wrapper from `master`, which no longer throws `URL Does Not Resolve` errors.
New view that validates the FHIR Validator Service Version against the expected version, and displays a warning if they don't match.
New README section with validator update instructions

## Testing guidance
Reset the validator version in `docker-compose.yml` to something before `1.2.0` (for example, `1.1.0`), run `docker-compose pull` and `docker-compose up --force-recreate`, and ensure that Inferno Program warns you on the homepage that you're not using the right validator version. The link won't work until this PR is merged however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,21 @@ all of which are hereby incorporated by reference.
       Nucleic Acids Res. 2004 Jan 1;32(Database issue):D267-70. doi: 10.1093/nar/gkh061.
       PubMed PMID: 14681409; PubMed Central PMCID: PMC308795.
 
-### Reference Implementation
+## Validator Service
+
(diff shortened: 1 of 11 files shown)


yunwwang (approved):
tested. looks good

arscan:
I pushed commit https://github.com/onc-healthit/inferno-program/pull/246/commits/0d8240e2bcbbe4cb3aecca37f7ec67d77de5f2ae to make the experience better if you are pointing to a very old validator.  Could you verify that this is ok @okeefm?  I verified this looks right on v.0.1.1 and v1.1.0.  See below for what it looks like with v0.1.1.
Screen Shot 2021-03-09 at 12 00 40 PM

arscan (approved):

okeefm:
@arscan yeah, that update looks good to me. I can't approve, because I'm a PR author, but I'll merge.

PR #247, opened 2021-03-04
Added PractitionerRole as prepopulated scope for UI

hershilpatel:
added PractictionerRole as prepopulated scope for the standalone patient app, limited app, and ehr practitioner app

# Summary
This scope was missing from default scopes causing 

## New behavior
Practitioner role scope is populated, USCPROV-04 passes on reference server

### DIFF
--- a/lib/app/views/onc_program.erb
+++ b/lib/app/views/onc_program.erb
@@ -560,7 +560,7 @@
               label: 'EHR-launched Scope',
               description: 'OAuth 2.0 scope provided by system to enable all required functionality',
               instance: instance,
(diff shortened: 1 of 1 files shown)


arscan:
So this is a weird one.  We do not include PractitionerRole in the default scopes on purpose because you do not need to support it.  But if you happen to have references to PractitionerRole, our scenario requires you to be authorized to them.  Our reference data set happens to have references to PractitionerRole, which means our reference server needs that scope.  So what we do is in the preset, we set that scope for the reference server.  But if you hit the reference server without using the preset, you need to know that you need to add that scope manually.

So, long story short, I don't think we want to do this because it implies that we expect systems to support this scope, when they don't have to.  It just turns out that our reference server (legally) does support it, and (legally) requires it to be granted.  So we put that in the preset.  And if you don't use the preset make sure you update that scope (just like how you have to update the client secret, etc).

Do we advertise what scopes are required on the reference server landing page?  I think that would be the appropriate place to put the scopes that work, perhaps?  It would be a long ugly list, because there are both user and patient variants.

arscan (changes requested):
Thanks for proactively addressing this, but per my above comment I don't think we want to do it.  If you agree, please close, otherwise let's talk about this one!

PR #250, opened 2021-03-09
FI-1152 Force existence of validators directory.

arscan:
# Summary

On a clean clone of Inferno, the /resources/terminology/validators directory does not exist.  This causes the following error:

```sh
> git clone git@github.com:/onc-healthit/inferno-program
> cd inferno-program
> docker-compose up --build

Creating clone-test_inferno_1           ... error

ERROR: for clone-test_inferno_1  Cannot create container for service inferno: invalid mount config for type "bind": bind source path does not exist: /Users/rscanlon/Projects/clone-test/resources/terminology/validators

ERROR: for inferno  Cannot create container for service inferno: invalid mount config for type "bind": bind source path does not exist: /Users/rscanlon/Projects/clone-test/resources/terminology/validators
ERROR: Encountered errors while bringing up the project.

```

## New behavior

This should allow a fresh copy of Inferno to run with `doccker-compose` out of the box, though they will still receive a warning about missing validators.

## Code changes

Added a .gitkeep file to the appropriate directory.

## Testing guidance

Clone the repository and run `docker-compose up --build`.  It should launch Inferno, and provide a warning about missing validators in the web interface.


okeefm (approved):
Tested, does what it should.

PR #259, opened 2021-03-18
FI-1182 Enable use of postgres db by configuration

arscan:
# Summary

Our use of an in-memory db makes it very difficult to scale Inferno as a mutli-user service.  This purpose of this PR is to allow users to configure Inferno to use a different db (e.g. Postgres).

## New behavior

In our docker environment, you can now configure inferno to use a different connection string to connect to a running postgres instance.  I created a docker-compose file that adds a postgres db that specifically uses this:

```
docker-compose -f docker-compose.production.yml up --build
```

This will primarily be used by us in the inferno.healthit.gov repo, but I added this other docker-compose file so others can do this without looking at the other repo.  I also updated the README to describe users how to use it if they'd like: https://github.com/onc-healthit/inferno-program/tree/postgres#deploying-inferno-in-a-server-environment

## Code changes

The main thing that was changed in the application itself, besides accepting this different connection string, is I had to change many of the types in the database from String (which defaults to varchar(50)) to Text.  Inferno errors very hard if we try to save data that dosn't fit into the column, so I am very liberal with how much space we are giving the columns.  This most certainly isn't the most efficient thing to do be doing, but we won't have all that many rows in this database.  And whatever we are doing, it will be much better than SQLite here anyhow.

The other thing that is weird here is that we need to install another OS dependency to enable Datamapper access to the necessary header files to connect to postgres, otherwise the postgres adapter gem doesn't work.  The OS-level dependency *should* be present already on our images, but the only thing that I could get to work was to install `postgres-server-dev` package on ubuntu, (and not just client or client header files or whatever).  This seems like overkill, and I think it is a bug in Datamapper.  I'm worried forcing users to include this will make them install postgres on their local machine, even if they don't want to run Inferno with postgres... and it caused our CI tests to fail (see the failed runs in Github actions).  So I took the very annoying step of splitting out our docker's Gemfile/Gemfile.lock from the one you would use if just running it natively.  So if you are not using Docker, you don't get the postgres adapter.

I don't love this solution, because it will be a big pain to maintain two copies of Gemfile.locks.  But this is really intended to be a short term solution to stabalize production until we get our improved system in there.  If there is a better solution I'd love to hear it.

Note: this will be followed by an update to our `inferno.healthit.gov` docker-compose file.

Also Note: I updated the git workflow to test this production version explicitly.  This caused an error to surface where the wrong patient_id is returned on `@testing_instance.patient_id`.  I do not think this test is relevant in the Program Edition, but to be safe, I fixed it. The fix for this is similar to how we fixed a similar issue in RequestResponse returning the wrong order.  **Unfortunately, this means that users using SQLite will have to clear out their databases if we keep this, and we'll have to put that in the CHANGELOG prominently.** I also added an explicit test for request_response order, because I did not see that and felt like we should verify that worked properly when in postgres too.

## Testing guidance

* On the `main` branch, run `docker-compose up --build`, and run a bunch of tests.  This will help you ensure that users won't have to do anything special with existing SQLite files files when switching to version (with a slightly different schema).  NOTE: A LATER ISSUE UNCOVERED REQUIRED US TO MAKE A SCHEMA FIX THAT REQUIRES USERS TO CLEAR THEIR SQLITE DB UNFORTUNATELY.
* Switch to this branch
* Run the regular `docker-compose up --build`, which still uses SQLite, and ensure that the changes to the schema are compatible with the old SQLite schema by running the tests.  NOTE: THIS NO LONGER WORKS, PLEASE DELETE DATA IN DATA DIRECTORY.
* [Review the updated README](https://github.com/onc-healthit/inferno-program/tree/postgres#deploying-inferno-in-a-server-environment) for how to run the 'server deployment version', ensuring the directions make sense and you can run the full set of tests.
* Run multiple tests at once in different tabs, verifying stability of the new setup.


lib/app/models/testing_instance.rb
```diff
@@ -10,88 +10,88 @@ module Models
...
-      property :bulk_data_jwks, String
-      property :bulk_access_token, String
+      property :bulk_data_jwks, Text
+      property :bulk_access_token, Text
       property :bulk_lines_to_validate, String
-      property :bulk_status_output, String
-      property :bulk_patient_ids_in_group, String
-      property :bulk_device_types_in_group, String
+      property :bulk_status_output, Text
+      property :bulk_patient_ids_in_group, String, length: 500
```
yunwwang, line 86: any reason leave this as String instead of changing to Text?

arscan: Not really, I can update that too.

arscan:
@okeefm I added a volume and updated the instructions on how to delete it.  I also added a network so the two docker-compose files don't share the same network.

Question: should this be named "docker-compose.server.yml", to reflect that this is when you are running this on a shared server, instead of "docker-compose.production.yml", which maybe implies that we are stating that this is a complete "production" setup?  For example, production would want to be served over 80 or 443, probably?  I decided to name the section in the Readme "Deploying Inferno in a Server Environment" because I think that is the most clear way of describing "intended for multiple users on a shared server".  On the other hand, "production" is a common term used to describe a deployed server environment.

arscan:
FYI: I updated the git workflow to test this config, which caused an error to surface. This error requires users to clear out their databases anyhow if they are using SQLite unfortunately.

Jammjammjamm (approved):

okeefm (approved):
Tested, looks good!

PR #277, opened 2021-04-30
FI-1187 Update search by post tests.

arscan:
# Summary

This PR includes a single POST-method search in the first search test for each profile type.  Search by POST is required by the base R4 specification, even though it is not mentioned at all in US Core 3.1.1.  We are interpreting GET as a requirement of US Core v3.1.1, at least for this version of the g10 tests.

## New behavior

 Instead of reissuing every query as a POST, this update uses the same strategy as checking that searching by references support including the resource type.  It performs a single search with the POST variant in the first test of each resource type, and then checks to make sure that the Bundle returned has the same number of entries.  When inspecting the requests/responses in the UI, you can now see a single POST search in each of the first tests of the US Core resource tests.

## Code changes

This involved updating `fhir_models` to allow parameters to be sent in the body of the request (in earlier versions, you could only change the method and parameters were always sent via querystring, which is odd behavior).  The test generator (and unit test generator) for US core was updated to include the additional checks.  The test descriptions were also updated.

## Testing guidance

Run the full set of single patient tests against the reference server and inspect the results.  There should be POST requests in the first test in each sequence.  Check to make sure the headers etc are what you would expect.


generator/uscore/uscore_generator.rb
```diff
@@ -404,7 +403,25 @@ def create_search_test(sequence, search_param)
 
         is_first_search = search_param == find_first_search(sequence)
 
-        search_test[:description] += 'Because this is the first search of the sequence, resources in the response will be used for subsequent tests.' if is_first_search
+        if is_first_search
+          search_test[:description] += %(
+            This test will verifies that the server supports searching by
```
Jammjammjamm, line 408:
```suggestion
            This test verifies that the server supports searching by
```

Jammjammjamm (commented):
The request body is being displayed as a ruby hash rather than a url-encoded string for the post searches:

Screen Shot 2021-04-30 at 8 30 27 AM

arscan:
> The request body is being displayed as a ruby hash rather than a url-encoded string for the post searches:

I've known about this issue for awhile, but we haven't really gotten to addressing it because of how few POSTs we do through this test procedure.  Might as well fix it now though...

czh-orz (approved):
Looks good to me. Once the payload display thing is fixed, should be good.

arscan:
I updated our FHIR request logger so it logs request payload to the URL encoded if that makes sense.

Screen Shot 2021-05-03 at 2 02 17 PM

Works for our LoggedRestClient too:
Screen Shot 2021-05-03 at 1 50 27 PM

I removed a line of code that caused our request logger to hashify JSON for some reason.  My guess is we preferred the hash version... somewhere?  Maybe our JSON doesn't pretty-print in certain cases?

Jammjammjamm (approved):

PR #280, opened 2021-05-04
Allows additional value in response's content-type

yunwwang:
# Summary
Allow additional values in response's content-type header

## New behavior
After receiving response for file download request, Inferno checks if response's content-type header contains application/fhir+ndjson

## Code changes
bulk_data_group_validation_sequence: change string compare to string include
add unit test for content-type header with multiple values

### DIFF
--- a/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
@@ -106,7 +106,7 @@ module Inferno
         }
 
         streamed_ndjson_get(file['url'], headers) do |response, resource|
-          assert response.header['Content-Type'].include?('application/fhir+ndjson'), "Content type must have 'application/fhir+ndjson' but found '#{response.header['Content-type']}'"
+          assert response.header['Content-Type']&.include?('application/fhir+ndjson'), "Content type must have 'application/fhir+ndjson' but found '#{response.header['Content-type']}'"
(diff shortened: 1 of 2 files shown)


lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -106,7 +106,7 @@ def check_file_request(file, klass, validate_all = true, lines_to_validate = 0,
         }
 
         streamed_ndjson_get(file['url'], headers) do |response, resource|
-          assert response.header['Content-Type'] == 'application/fhir+ndjson', "Content type must be 'application/fhir+ndjson' but is '#{response.header['Content-type']}'"
+          assert response.header['Content-Type'].include?('application/fhir+ndjson'), "Content type must have 'application/fhir+ndjson' but found '#{response.header['Content-type']}'"
```
Jammjammjamm, line 109: It doesn't look like we've verified that the `Content-Type` header is present before this, so I think we need a nil-safe check:
```suggestion
          assert response.header['Content-Type']&.include?('application/fhir+ndjson'), "Content type must have 'application/fhir+ndjson' but found '#{response.header['Content-type']}'"
```

okeefm:
Agree with @Jammjammjamm on his comment; otherwise looks good.

Jammjammjamm (approved):

PR #282, opened 2021-05-07
FI-1215 Skip search with patient id if patient does not exist

yunwwang:
# Summary
Fix the failure on DiagnosticReport if patient does not exist

## New behavior
Code generator add line to skip search with patient id (except first search) if patient does not exist

## Code changes
uscore_generator.rb
add next if logic at get_search_params()
remove grab_first_value parameter from get_search_params() since it is not used by any caller

## Testing guidance
Add an no existing patient id to the input popup. All tests shall still pass

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1203,13 +1203,20 @@ module Inferno
           #{skip_if_not_found_code(sequence)})
       end
 
-      def get_search_params(search_parameters, sequence, grab_first_value = false)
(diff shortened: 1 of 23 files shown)


arscan (approved):
I tried out the pre-merge against the reference server with presets, except I added a non-existent patient id to the list, and failed as expected.  When I tried it with this branch, it passed as expected.

I took a look at the code and it seems like it was added in all the right tests.

The complexity of these tests is now at the point where I am a bit wary of any changes at all, but I think this is ok.

Jammjammjamm (approved):

PR #283, opened 2021-05-07
FI-1189 CCG Scope Updates

arscan:
# Summary

Addresses CCG updates regarding  the absence of a requirement for denying `offline_access` scope, and allowing systems to use `user` scopes for non-Patient compartment resources for patient app launches.  Note that this addresses both FI-1189 and FI-1188 in our internal tracker due to the close relationship of the two CCG clarifications.

## New behavior

For the 'limited access tests', there should be no mention of needing to deny "offline_access", and when you run the test it should pass if you allow "offline_access".

For the 'standalone patient app' tests, you can change the defaulted "patient" scope to "user" for non-patient compartment resources (and technically not encounter either since it is not uscdi v1).  Try changing some scopes to 'user', it should fail if you do it on a patient compartment resource (e.g. Condition), but pass if you do it on any/all non-patient compartment resources (e.g. Practitioner).

## Code changes

Changes the standalone patient app launch tests and the limited access tests.  Also updated the unit tests.

## Testing guidance

See 'new behavior' above.

### DIFF
--- a/generator/uscore/templates/access_verify_sequence.rb.erb
+++ b/generator/uscore/templates/access_verify_sequence.rb.erb
@@ -12,14 +12,13 @@ module Inferno
       test_id_prefix 'AVR'
       details %(
         This test ensures that patients are able to grant or deny access to a
-        subset of resources to an app. It also verifies that patients can
-        prevent issuance of a refresh token by denying the `offline_access`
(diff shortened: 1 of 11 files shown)


Jammjammjamm (approved):

okeefm (approved):
Looks good

Issue #281, opened 2021-05-06
MIME types with parameters not recognized as valid Attachment.contentType codes


lcmaas:
valid MIME types that include valid parameters such as charset are not recognized for DocumentReference.content.attachment.contentType

* Edition of inferno (Community or Program): 1.5.0P

**Steps to reproduce**
1. Return a DocumentReference resource with DocumentReference.content.attachment.contentType = 'text/plain; charset=UTF-8' during execution of Single Patient API tests.
2. The resource fails USCDR-12 with the following error: DocumentReference.content.attachment.contentType with code 'text/plain; charset=UTF-8' is not in http://hl7.org/fhir/ValueSet/mimetypes

Or manually with rake:
$ run inferno bundle exec rake "terminology:check_code[text/plain; charset=UTF-8,urn:ietf:bcp:13,http://hl7.org/fhir/ValueSet/mimetypes]"
returns: urn:ietf:bcp:13|text/plain; charset=UTF-8 is not in http://hl7.org/fhir/ValueSet/mimetypes

**Expected behavior**
MIME types with valid parameters should pass.

From the US Core definition for [contentType](http://hl7.org/fhir/us/core/STU3.1.1/StructureDefinition-us-core-documentreference-definitions.html#DocumentReference.content.attachment.contentType) (which is also the R4 Attachment.contentType definition):

> Definition: Identifies the type of the data in the attachment and allows a method to be chosen to interpret or render the data. **Includes mime type parameters such as charset where appropriate.**

Note that the this is an infinitely large value set, including every combination of permitted parameter and possible value for every valid MIME type and subtype.

**Actual behavior**
test fails with message: DocumentReference.content.attachment.contentType with code 'text/plain; charset=UTF-8' is not in http://hl7.org/fhir/ValueSet/mimetypes

yunwwang:
Hello @lcmaas, this issue is fixed in Inferno program release 1.6.1.

PR #291, opened 2021-05-13
Add support for "preprocessing" BCP:13 codes prior to validation

okeefm:
# Summary
BCP:13 codes (aka `MIME Types`) can have optional "parameters" tacked on after the actual MIME type; like `application/fhir+json; charset=UTF-8`. This is considered "legal" in FHIR, but our Bloom Filters can't handle them, so we have to strip off anything that's not the MIME type itself before validation.

While we're at it, this PR also `downcase`s the MIME type before validation, to avoid case mismatches (since MIIME types are case-insensitive). 

The PR also adds this preprocess step to the BCP:13 bloom filter generation code, so moving forward all the MIME types coming out of the MIME::types gem will be `downcase`d.

## Testing guidance
* Ensure the bloom filters are built/present on your system
* Test a MIME type with parameters against the appropriate valueset, with a Rake task like `rake terminology:check_code['application/fhir+json; CHARSET=utf-8','urn:ietf:bcp:13','http://hl7.org/fhir/ValueSet/mimetypes']`
* Optional: run the same task, but change the MIME type to be gibberish, and ensure the code check fails
* Optional: run the terminology build tasks (after deleting the bloom filters) and test a code like `application/A2L` (which is upcased coming from the `mime-types` gem) and ensure it passes both upcase'd and downcase'd

### DIFF
--- a/lib/app/utils/bcp_13.rb
+++ b/lib/app/utils/bcp_13.rb
@@ -13,9 +13,15 @@ module Inferno
         require 'mime/types'
         cs_set = Set.new
         MIME::Types.each do |type|
-          cs_set.add(system: 'urn:ietf:bcp:13', code: type.simplified)
+          cs_set.add(system: 'urn:ietf:bcp:13', code: preprocess_code(type.simplified))
         end
         cs_set
+      end
+
+      # "preprocess" step for BCP13 codes, to remove any of the optional parameters
+      # and downcase them
+      def self.preprocess_code(code)
+        code&.split(';')&.first&.downcase
       end
     end
   end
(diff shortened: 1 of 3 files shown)


radamson (approved):
Code looks good and everything worked as expected testing locally.

Do we need to port this to `onc-healthit/inferno` too?

okeefm:
@radamson Probably not. inferno-community is delegating all of its terminology to tx.fhir.org with a PR that's open now.

okeefm:
Fixes #281

arscan (approved):
Looks good!

PR #296, opened 2021-05-26
read config file from rake tests to set disable verify peer correctly

hershilpatel:
# Summary
Updating running tests from tasks.rake so it reads the config file

## New behavior
Now tasks.rake manually reads the config file to learn about verify_peer

## Testing guidance
Change the config's disable_verify_peer Run inferno tests from rake to test and confirm that USCCAP-01 is ommited

### DIFF
--- a/lib/tasks/tasks.rake
+++ b/lib/tasks/tasks.rake
@@ -7,7 +7,6 @@ require 'csv'
 require 'colorize'
 require 'optparse'
 require 'rubocop/rake_task'
-require 'yaml'
 
 require_relative '../app'
 require_relative '../app/endpoint'
@@ -78,10 +77,7 @@ def execute(instance, sequences)
     end
     instance.save!
 
-    # manually read config yaml file
-    config = YAML.load_file(ENV['INFERNO_CONFIG_FILE'] || 'config.yml')
-    disable_verify_peer = config['disable_verify_peer']
-
+    disable_verify_peer = Inferno::App::Endpoint.settings.disable_verify_peer
     sequence_instance = sequence.new(instance, client, disable_verify_peer)
     sequence_result = nil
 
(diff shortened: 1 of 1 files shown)


lib/tasks/tasks.rake
```diff
@@ -76,7 +77,12 @@ def execute(instance, sequences)
       end
     end
     instance.save!
-    sequence_instance = sequence.new(instance, client, false)
+
+    # manually read config yaml file
+    config = YAML.load_file(ENV['INFERNO_CONFIG_FILE'] || 'config.yml')
+    disable_verify_peer = config['disable_verify_peer']
```
arscan, line 83: Manually reading config file seems brittle.  I did some searching around how to leverage the work Sinatra is already doing here and came across a [SO question/answer with an example](https://stackoverflow.com/questions/36507763/how-to-access-sinatra-custom-settings-from-service) which led me to:

```
Inferno::App::Endpoint.settings.disable_verify_peer
```
That seems to work for me.  Have you tried this?

arscan (changes requested):

arscan:
I came across a pretty easy way to leverage Sinatra's built in config parsing, instead of separately re-reading the config file as is done here.  Since @hershilpatel is out this week, I've made the change.  FYI @okeefm -- do you see any downside to this alternate approach?

okeefm (approved):
Works for me; if I change the value in `config.yml`, I see it change in `tasks.rake` if I pry into it.

okeefm:
I can't think of any particular downsides. I think the code in `Inferno::App::Endpoint` is already getting executed either way, since we require it, so it's not like there's any side-effects that weren't already happening.

arscan (approved):

PR #300, opened 2021-06-04
Update changelog, version, and matrix.

arscan:
The **Inferno Program Edition v1.6.1** release fixes one test issue that may incorrectly cause systems to fail tests. It also includes a number of fixes not directly related to test execution. Systems that passed tests in Inferno Program Edition v1.6.0 should continue to pass in v1.6.1.

Updates in this version include:

* Fixes issue where BCP:13 code validation incorrectly fails systems that include optional MIME Type parameters, such as "charset=UTF-8".
* Fixes issue introduced in v1.6.0 that prevents new installations of Inferno from generating the necessary terminology files used for terminology validation.
* Fixes issue that prevented the command line interface from disabling TLS Peer Validation, which often isn't necessary for CI-related tests.


okeefm (approved):
Proofread the PR text and the text in the ERB template. Ensured the Excel file will open, but did not verify accuracy of every cell. Looks good.

PR #305, opened 2021-06-14
FI-1192: Fix search calls

Jammjammjamm:
# Summary
The access verification sequences were passing a string instead of a fhir model to `FHIR::Client#search`, which was causing errors to be logged.

## Code changes
Fixed the template for these sequences to use a `FHIR::Model` class rather than a string in their search calls. Also, #283 added some code directly to a generated sequence, so I moved it into the template.

## Testing guidance
When running the verify unrestricted access sequence, you shouldn't see these error messages in the logs:
```
ERROR | Inferno | Failed to parse application/fhir+json as resource Procedure: undefined method `ancestors' for "Procedure":String
```

### DIFF
--- a/generator/uscore/templates/access_verify_sequence.rb.erb
+++ b/generator/uscore/templates/access_verify_sequence.rb.erb
@@ -115,8 +115,11 @@ module Inferno
       end
 
       def scope_granting_access(resource, scopes)
-        scopes.split(' ').find do |scope| 
-          ['patient/*.read', 'patient/*.*', "patient/#{resource}.read", "patient/#{resource}.*"].include? scope
+        non_patient_compartment_resources = ['Encounter', 'Device', 'Location', 'Medication', 'Organization',
+          'Practitioner', 'PractitionerRole', 'RelatedPerson']
+        scopes.split(' ').find do |scope|
+          return true if non_patient_compartment_resources.include?(resource) && ["user/#{resource}.read", "user/#{resource}.*"].include?(scope)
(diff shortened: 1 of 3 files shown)


generator/uscore/templates/access_verify_sequence.rb.erb
```diff
@@ -115,8 +115,11 @@ module Inferno
       end
 
       def scope_granting_access(resource, scopes)
-        scopes.split(' ').find do |scope| 
-          ['patient/*.read', 'patient/*.*', "patient/#{resource}.read", "patient/#{resource}.*"].include? scope
+        non_patient_compartment_resources = ['Encounter', 'Device', 'Location', 'Medication', 'Organization',
+          'Practitioner', 'PractitionerRole', 'RelatedPerson']
+        scopes.split(' ').find do |scope|
+          return true if non_patient_compartment_resources.include?(resource) && ["user/#{resource}.read", "user/#{resource}.*"].include?(scope)
+          ['patient/*.read', 'patient/*.*', "patient/#{resource}.read", "patient/#{resource}.*"].include?(scope)
```
arscan, line 122: Ah, good catch, I find it surprising that this happened -- this must be the first time we've regenerated tests since I made this update.  I looked through the code and it's a situation where `non_patient_compartment_resources` elements will never be passed into this method in the restricted sequence, so there should be no effect to having this code in here (except for generation to not clobber the unrestricted sequence, where we need it).

arscan (approved):
Looks good.  Tried out both single patient launch sequences and didn't see the errors.  Also tried a few situations where I knew it would fail (and it did properly).

okeefm (approved):
Tested; no longer see those messages when running the unrestricted access verify sequence.

PR #308, opened 2021-06-24
Move negative authorization code tests to its own sequence

yunwwang:
# Summary
Some servers may revoke authorization code after negative test which blocks following tests in authorization test sequence. To avoid that, we moved negative authorization code tests to its own sequence as an additional test in "other" tab.

## New behavior
Replace two negative authorization code tests in ONC EHR Launch Sequece, ONC Standalone Launch Sequence, and ONC Standalone Restricted Launch Sequence 

Add Invalid Authorization Code Sequence

### DIFF
--- a/lib/modules/onc_program/smart_invalid_authorization_code_sequence.rb
+++ b/lib/modules/onc_program/smart_invalid_authorization_code_sequence.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+require_relative './shared_onc_launch_tests'
+
+module Inferno
+  module Sequence
+    class SMARTInvalidAuthorizationCodeSequence < SequenceBase
+      include Inferno::Sequence::SharedONCLaunchTests
+
+      title 'SMART App Launch Error: Invalid Authorization Code'
+      description 'Demonstrate that the server properly validates Authorization code'
+
+      test_id_prefix 'SIAC'
+
+      requires :onc_sl_client_id,
+               :onc_sl_confidential_client,
+               :onc_sl_client_secret,
+               :onc_sl_scopes,
+               :oauth_authorize_endpoint,
+               :oauth_token_endpoint,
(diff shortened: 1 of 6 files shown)


lib/modules/onc_program/smart_invalid_authorization_code_sequence.rb
```diff
@@ -0,0 +1,107 @@
...
+        oauth2_params.each do |key, value|
+          oauth2_auth_query += "#{key}=#{CGI.escape(value)}&"
+        end
+
+        redirect oauth2_auth_query[0..-2], 'redirect'
+      end
+
+      code_and_state_received_test(index: '02')
+      invalid_code_test(index: '03')
+      invalid_client_id_test(index: '04')
```
Jammjammjamm, line 104: Do the invalid code and client_id tests need to be split up into their own sequences?

Jammjammjamm (commented):
This is a nice solution.

lib/modules/onc_program/smart_invalid_authorization_code_sequence.rb
yunwwang: It is not necessary. The issue this ticket trying to address is that the authorization code can be used only once. The "invalid_code_test" sends request with an faked/invalid authorization code which does not affect the actual/valid authorization code. The second test "invalid_client_id_test" sends the request using the valid authorization code with invalid client id which may cause revocation of such code by server. 
Actually, it is not required to move the "invalid_code_test" to this stand alone sequence since it does not affect the authorization code as explained. Rob and I decided to move it in order to grouping related negative tests together.

Jammjammjamm (approved):

okeefm (approved):
Tested. Invalid auth code sequence seems to work, and deprecated tests are not showing up in the UI. Looks good

PR #311, opened 2021-06-28
Fi 1262 search string starts with

czh-orz:
# Summary
This addresses an issue where tests that searched against a string target value incorrectly failed. This happens because the validation checked if the searched value was the value in the resource. It should instead just check if the value in the resource starts with the searched value. https://www.hl7.org/fhir/search.html#string

Eg.) searching for family="test" might return two patients with family="test" and family="testing". The "testing" value would've incorrectly failed validation. 

## New behavior
Searches against string values will check that the value starts with the searched value.

## Code changes
I changed the validation portion of validate_resource_item for parameters that had a string search value.

### DIFF
--- a/test/unit/sequence_base_test.rb
+++ b/test/unit/sequence_base_test.rb
@@ -58,6 +58,32 @@ class SequenceBaseTest < MiniTest::Test
         end
       end
     end
+
+    it 'passes when string value starts with searched value' do
+      @instance = Inferno::TestingInstance.create!
+      client = FHIR::Client.new('')
+      patient_sequence = Inferno::Sequence::USCore311PatientSequence.new(@instance, client, true)
+      patient_resource = FHIR::Patient.new(
+        name: [{
+          family: 'LastName'
+        }]
+      )
+      patient_sequence.validate_resource_item(patient_resource, 'family', 'las')
+    end
+
+    it 'fails when string value does not start with searched value' do
+      @instance = Inferno::TestingInstance.create!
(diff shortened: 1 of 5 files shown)


Jammjammjamm (approved):

test/unit/sequence_base_test.rb
```diff
@@ -58,6 +58,32 @@ class SequenceBaseTest < MiniTest::Test
         end
       end
     end
+
+    it 'passes when string value starts with searched value' do
```
okeefm, line 62: Should these tests be in the `US Core Patient Test`? Or is that the wrong place for them (either because they're autogenerated, or another reason?)

arscan: I agree that it is very odd to have a dependency on USCore311 tests here, and ideally the code would be written such that you could wouldn't need a dependency on a specific test.  What if we remove the test.  Having said that, all tests in this entire file have this problem, the addition adds useful test coverage, and we're planning on migrating off this platform anyhow in the future... so a big refactor to fix this doesn't make sense to me.

arscan:
I spent awhile looking at how our `name` tests behave, and as part of that I added a few tests to make sure that it all worked as intendend.  I did notice that we do not support fuzzy search by name, which is something we should support, as it is required by the spec. One thing that is problematic is that technically FHIR also allows servers to add more results if they 'believe' them to be useful, so we may want to figure out how to allow that behavior without throwing out all this checking we do.

arscan (approved):

PR #322, opened 2021-07-14
Updated Provenance Test to not require search

hershilpatel:
# Summary
USCPROV-04 fails on the reference server because the test required Provenance search. The reference server conformance statement do not support search as US Core does not require it. Updated the test to only check for read.

## New behavior
The reference server should pass all Single Patient Api tests (specifically USCPROV-04)


hershilpatel:
This needs to be fixed in the generator, so closing

PR #323, opened 2021-07-21
Fi 1269 reference server pass USCPROV 04

hershilpatel:
# Summary
The inferno-reference-server fails test USCPROV 04 because it requires search to be supported. This fails because reference server's capability statement does not declare for support for Provenance search

## New behavior
The generator was changed to not check for read and support being declared in cabablity statement, as it is checked in previous tests. The the tests were regenerated with this in mind.

## Testing guidance
Run the reference server through this version of inferno and USCPROV-04

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -716,7 +716,7 @@ module Inferno
       def create_resource_profile_test(sequence)
         test_key = :validate_resources
         test = {
-          tests_that: "#{sequence[:resource]} resources returned from previous search conform to the #{sequence[:profile_name]}.",
+          tests_that: "#{sequence[:resource]} resources returned conform to the #{sequence[:profile_name]}.",
           key: test_key,
           index: sequence[:tests].length + 1,
           link: sequence[:profile],
@@ -924,7 +924,6 @@ module Inferno
 
         resource_array = sequence[:delayed_sequence] ? "@#{sequence[:resource].underscore}_ary" : "@#{sequence[:resource].underscore}_ary&.values&.flatten"
         test[:test_code] = %(
-              skip_if_known_not_supported(:#{sequence[:resource]}, [:search, :read])
               #{skip_if_not_found_code(sequence)}
 
               validated_resources = Set.new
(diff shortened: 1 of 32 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -716,7 +716,7 @@ def structure_to_string(struct)
       def create_resource_profile_test(sequence)
         test_key = :validate_resources
         test = {
-          tests_that: "#{sequence[:resource]} resources returned from previous search conform to the #{sequence[:profile_name]}.",
+          tests_that: "#{sequence[:resource]} resources returned conform to the #{sequence[:profile_name]}.",
```
Jammjammjamm, line 719: I think we should run this wording change by @arscan.

arscan: I think I'm ok with it.  For these, try to put yourself in the position of a developer that is looking at these tests for the first time.  If they fail the test, will they ask "what resources returned where?"  I would like to train people to be reading the test details if it is hard to be succinct here, so maybe that's the answer (we beef up the details section).  Or maybe we can just state "...resources received during previous tests...".

arscan: Now that we are really getting into the wording here, this isn't quite accurate.  Sometimes we only look at SOME of the resources, because not all Observation resource types need to conform to the BodyWeight profile, for example.  Should we instead say "Applicable #{sequence[:resource]} resources received during previous tests"?

hershilpatel: I like that wording, still succinct but more descriptive, I'll make the update

arscan: I was looking at the other tests, and the downside of rewording it too much is that then the wording may start misaligning with the language in the other tests, so we get into the situation where we are renaming all tests.  I think maybe just going with "returned during previous tests" is good and we'll leave it at that.  Thansk!

arscan (approved):
The slightly weird thing about this is that if you don't properly do your conformance statement, you'll get a mix of skip messages... some tests say something like 'you don't say you support this resource' while some will say 'you haven't returned any resources to us'.  Turns out we did this already a little with the 'Must Support' tests, so this isn't really making it any worse from that perspective, and it certainly is fixing the issue we had before.  So I think this is fine to merge.

Issue #324, opened 2021-07-22
Revoked Refreshed Token input field on the Token Revocation screen is not populated (see attachment)


IgLysk00:
Thank you for reporting a possible bug in Inferno! Please fill in as much of the template below as you can.

**Subject of the issue**
See Issue description

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 1.6.1
* Which browser and version(s) is the bug present on?: Chrome

**Steps to reproduce**
Inferno installed on AWS. Create your AWS Inferno instance. Use Program Edition. Follow the Drummond workflow – 
Standalone Patient App, 
Standalone Patient App – Limited Access, 
Other - Public Client Standalone Launch with OpenID Connect, Token Revocation. Start running Token Revocation test.

Include:

* The selected test suite
* The URL of the FHIR server being tested
* The Test ID, if it's a problem with a specific test.
* Any required configuration options (client ID, client secret, etc.)
* Anything else needed to make the issue occur.

**Expected behavior**
The Revoked Refreshed Token field should be populated

**Actual behavior**
![image](https://user-images.githubusercontent.com/73243391/126670742-a7bcf300-af04-4c4d-9fc4-c6cf97d37b6d.png)

arscan:
Thanks for the report!  This does appear to be an issue with how we ordered some of the tests, if the expectation is that all these fields will be prefilled so the tester doesn't need to type them in (which is a valid expectation!).  The last launch that occurred needs to have been granted the `offline_access` or `online_access` scope, because we are prefilling these fields based on data from the last launch.  The 'Public Client Standalone Launch with OpenID Connect' does not use those scopes, therefore shouldn't return a refresh token, and therefore that field won't be prepopulated if run right after that launch.

Could you try this sequence, but instead run `Token Revocation` prior to the `Public Client Standalone Launch with OpenID Connect`?  That way, it will prefill the field with the refresh token that should have been granted during the limited access launch, which now includes the 'offline_access' scope?  If that works, we should reorder these in the interface, or at least add some clarifying language stating that 

This has also exposed what looks to be a deficiency in our reference server's mocked SMART interface too, which we are looking into.

IgLysk00:
Tried to execute Token Revocation prior to Public Client Standalone Launch with OpenID. The result is the same - the input field on the screen is not populated.

[...]

PR #325, opened 2021-07-26
FI-1287 Save bearer and refresh tokens to separate state variable in 1st launch

arscan:
# Summary

Inferno's sequences often overwrite certain shared state variables, which can be problematic if you want to have a later sequence use output from a specific instance of sequence prior (instead of just the last one that was run).  To work around this, in certain cases we save some of these variables under a separate name in a derived sequence, in the case where we do NOT want them overwritten.  For example, we save the bearer/refresh token that comes out of the 'Standalone Patient App' into a separate variables, because we want to revoke those specific tokens at the very end of the procedure (Other > Token Revocation) and don't want the other launches to overwrite those values.  

This fixes an issue that prevented the refresh token from being saved in a separate variable in a certain case, which caused it to not show up as a default input in the variable. See #324

It also fixed an issue where the onc_sl_patient_id wasn't being displayed as an input in the Token Revocation test, because we misnamed the input field in the .erb file.  We just silently grabbed the patient_id provided in the first Single Patient API tests without reporting it to the user.

This PR also just cleans up the code slightly here to make it a little more logical.  The end result is the same, but the values are saved and reported to the user in more logical spots.

Finally, and this tangential to the scope of this ticket, but it adds a (almost always invisible) input field for a Bearer Token for Bulk API that will show up in the developer knows how to disable authorization testing in the module file.  This was done because I was already editing the input parameter fields.

## New behavior

'Standalone Patient App > Standalone Launch WIth Patient Scope > Outputs' now includes "Onc Sl Token', 'Onc Sl Refresh Token', and 'Onc Sl Patient Id' outputs.  Before it does not include those values, because it relied on the 'Token Refresh' test to do this, which we shouldn't do.

'Standalone Patient App > Discovery' now is explicit that is saving the "onc_sl_*" for the auth endpoints.  It was doing this before, but we didn't report it as 'output'.

The Token Revocation sequence now has an input for patient_id, which should be prefilled from the context provided in the first launch (as is the bearer and refresh tokens).

## Code changes

Updated the Standalone Launch with Patient Scopes to save that data to those other variables for safekeeping.  Note that this is basically just copying a few lines from our ONC Standalone Launch Refresh Token sequence.  I also slightly reorganized the code to be more clean.

## Testing guidance

To truly test this fixes a real life issue, you'd have to alter our reference implementation token to not providing a 'refresh_token' parameter when a token is refreshed.  I don't think that is necessary for here though, just check that 'Standalone Patient App > Standalone Launch WIth Patient Scope > Outputs' now includes "Onc Sl Token' and 'Onc Sl Refresh Token' outputs.  Also verify that if you the 'Single Patient API' launch, followed by different tests that involve launches (e.g. Limited App Launch, EHR Launch, or Other > Public Client launch), that the Token Revocation only shows the Bearer/Refresh Tokens/Patient IDs from the 'Single Patient API' launch and was not overwritten.

### DIFF
--- a/lib/app/views/onc_program.erb
+++ b/lib/app/views/onc_program.erb
@@ -493,11 +493,11 @@
               value: instance.onc_sl_refresh_token,
               })%> 
 
-        <%= erb(:prerequisite_field,{},{prerequisite: :patient_id,
+        <%= erb(:prerequisite_field,{},{prerequisite: :onc_sl_patient_id,
               label: 'Patient ID',
-              description: 'Patient ID provided as context in the patient standalone launch.',
+              description: 'Patient ID associated with revoked tokens provided as context in the patient standalone launch. This will be used to verify access is no longer granted using the revoked token.',
               instance: instance,
-              value: instance.patient_id,
+              value: instance.onc_sl_patient_id,
               })%>
 
         <%= erb(:prerequisite_field,{},{prerequisite: :dynamic_registration_token,
@@ -654,7 +654,12 @@
               instance: instance,
               value: instance.bulk_token_endpoint,
               })%>       
-
+        <%= erb(:prerequisite_field,{},{prerequisite: :bulk_access_token,
+              label: 'Bulk Data Access Token',
+              description: 'Bearer token to be used for Bulk Data API.  Note that this field is only intended to support development for systems that have not yet implemented authorization and must not be available as input for certification.',
+              instance: instance,
+              value: instance.bulk_access_token,
+              })%> 
         <%= erb(:prerequisite_field,{},{prerequisite: :bulk_client_id,
               label: 'Bulk Data Client ID',
               description: 'Client ID provided at registration to the Inferno application.',
--- a/lib/modules/onc_program/onc_standalone_launch_sequence.rb
+++ b/lib/modules/onc_program/onc_standalone_launch_sequence.rb
@@ -24,7 +24,7 @@ module Inferno
                :initiate_login_uri,
                :redirect_uris
 
-      defines :token, :id_token, :refresh_token, :patient_id
+      defines :token, :id_token, :refresh_token, :patient_id, :onc_sl_token, :onc_sl_refresh_token
 
       show_uris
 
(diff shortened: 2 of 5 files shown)


lib/modules/onc_program/onc_standalone_launch_sequence.rb
```diff
@@ -109,6 +109,35 @@ def instance_scopes
...
+        @instance.save!
+      end
+
+      def after_save_access_token(token)
+        # See `after_save_refresh_token` method for explanation of purpose of
+        # this method.
+
+        @instance.onc_sl_token = token
+
+        # save a copy so patient_id and oauth_token_endpoint are not overwritten
```
Jammjammjamm, line 135: I don't see `oauth_token_endpoint` being saved.

lib/app/views/onc_program.erb
```diff
@@ -654,7 +654,12 @@
               instance: instance,
               value: instance.bulk_token_endpoint,
               })%>       
-
+        <%= erb(:prerequisite_field,{},{prerequisite: :bulk_access_token,
+              label: 'Bulk Data Access Token',
+              description: 'Bearer token to be used for Bulk Data API.  Note that this field is only intended to support development for systems that have not yet implemented authorization and must not be available as input for certification.',
+              instance: instance,
+              value: instance.bulk_access_token,
+              })%>
```
arscan, line 662: Note: this is out-of-scope for this ticket technically, but we've found it is really useful to have it so that developers can bypass the Bulk Auth while developing the Bulk API if auth isn't yet implemented.  I left it in because I fixed other inputs and was touching this file already.

lib/modules/onc_program/onc_standalone_launch_sequence.rb
arscan: This made me nervous, so I reviewed how we are saving fields 'off to the side'.  I found that we were a little sloppy with reporting when certain things were saved, and where they were saved, particularly from an 'Output' tab perspective.  I went through and cleaned it up all.  I think, when we wrote this all, we weren't expecting that prefilled data in the 'Token Refresh' tab would be relied on quite so heavily, so having it work was good enough.  But now it is clear that is a bit more than just convenience for the tester, because they get confused when it isn't filled.

I recommend looknig again from the start, since I've made a lot more changes.

Jammjammjamm (approved):

yunwwang (approved):
Tested. Everything works as expected

PR #334, opened 2021-08-04
FI-1292/1293 Add missing index and allow dynamic content in db config

Jammjammjamm:
This branch adds the missing index to join table to fix performance issues as the table size increases.

Also allows erb to be used in the database configuration so that values can be loaded from environment variables.

## Testing guidance
This should work:
```
docker-compose -f docker-compose.postgres.yml build
docker-compose -f docker-compose.postgres.yml up
```
You can validate that the environment variables work by adding something like this to the inferno service in `docker-compose.postgres.yml`, which will cause inferno to fail to connect to the database.
```
    environment:
      INFERNO_DB_USERNAME: xyz
```

### DIFF
--- a/db/config.postgres.yml
+++ b/db/config.postgres.yml
@@ -1,3 +1,6 @@
+# ERB may be used to add dynamic content to this file.
+# https://docs.ruby-lang.org/en/2.7.0/ERB.html
+
 development:
   adapter: postgresql
   database: inferno_program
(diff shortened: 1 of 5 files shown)


yunwwang (approved):

arscan (approved):
Looks good.  I checked:
* for postgres config, i manually verified the index was created
* for postgres config, app launched and worked without any issues
* for postgres config, if i altered the username by environment variable, it does indeed fail
* for sqlite config, the migration appeared to work (didn't manually verify index though)
* for sqlite config, the app worked without issues

Issue #235, opened 2021-02-12
Support for IE 11 in EHR Launch Test (FI-1156)


dougpratt:
_**Is your feature request related to a problem? Please describe it.**
Describe the problem you are trying to solve._

Our EHR requires Internet Explorer 11. When we run the INFERNO test for EHR Launch, INFERNO fails with a message that it does not support Internet Explorer (see attachment 1). When we click on OK, and then click on the Run Tests button, nothing happens (see attachment 2). I asked about this during the monthly call (Feb 10 @1 PM EST), and Rob Scanlon asked me to log this request.

![Inferno-1](https://user-images.githubusercontent.com/27502666/107796136-848e4300-6d27-11eb-9393-62cd6610258c.jpg)
![Inferno-2](https://user-images.githubusercontent.com/27502666/107796148-88ba6080-6d27-11eb-88c8-9c6c923bff20.jpg)

_**Describe the solution you'd like to see implemented**_

Allow EHR Launch from IE 11. 

The retirement date announced by Microsoft is the same as Windows 10 retirement, which is 10/24/2025. See https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge and https://docs.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro.

_**Describe alternatives you've considered**
Describe any alternative solutions or features you have considered._
We have considered support for additional browsers, but will be certifying our EHR before supporting additional browsers.

arscan:
Thanks for letting us know about this.  We have no grounds to disallow usage of IE11 for the app launch, so we need to make sure that you can execute the launch process parts of those tests within the browser of your choosing.  Is it possible for you to use another browser to get Inferno into the state where it is expecting the launch, then only use IE11 for the EHR launch part, then switch back to the non-IE11 browser?  The following steps describe that process:

1) In a browser other than IE 11 (preferably chrome or firefox), initiate the EHR launch test sequence.  This essentially will put Inferno in a state where it is waiting for the EHR to start the app launch process.  You'll be presented with this popup:
![Screen Shot 2021-02-15 at 4 35 22 PM](https://user-images.githubusercontent.com/412901/107995069-f70f5500-6fab-11eb-8638-8fc4d2bbaa08.png)
2) From within the EHR, launch Inferno at that launch URI.  Click through the 'Inferno does not support IE...' alert box, and you should see the following:
[...]

PR #335, opened 2021-08-05
FI-1156 Fix issues in IE11 and slightly clean up certain alert message styling.

arscan:
# Summary
IE11 is not technically end-of-life yet and may be used for embedded apps within EHRs until it is.  You have always been able to work around this using a modern browser to start the launch test, launching within the embedded IE11 browser, then switching back to the modern browser & refreshing (see #235).  This update fixes javascript/styling to allow IE11 to use Inferno, which makes it clear that the certification criteria does not specifically disallow IE 11.

## New behavior

IE11 no longer has an alert message stating it is not supported.  IE10 and before should still have this alert message.  IE11 should look exactly the same as chrome/firefox/safari.

I also slightly cleaned up the alert message when you do not have terminology installed, or are pointing to the wrong version of the validator service.

## Code changes
I fixed the javascript (adding polyfills when needed) and CSS so that IE11 functions properly. I also updated the browser checking to allow IE11, but it should still pop an alert if you are using < IE11.

Also, I had to bump the inferno version to "1.6.3-pre" to bust caches on javascript/css.  Since we haven't yet merged any breaking changed into `main`, I didn't bump it to 1.7.0-pre, though it does look like we might end up including updates requiring us to go to 1.7.0.
## Testing guidance

Walk through the majority of the tests in chrome or firefox to make sure I didn't inadvertently break anything in those browsers.

I already did this (see comment below), but to be more thorough, download an IE11 VM from MS (or hop on a windows machine that has it) and try running through the tests (NOTE: our reference server's OAuth screens are not compatible with IE11, so you have to work around that a bit, but that is outside the scope of this ticket).

I also already did this (see comment below), but to be most thorough, also download IE10 in a VM to ensure that the error message still shows up.

Capture

### DIFF
--- a/lib/app/views/layout.erb
+++ b/lib/app/views/layout.erb
@@ -35,10 +35,6 @@
         </div>
       </nav>
     <% end %>
-    <script type="text/javascript">
-      if(/MSIE \d|Trident.*rv:/.test(navigator.userAgent))
-        alert('Inferno does not currently support Internet Explorer.  Please switch to another browser, such as Chrome, Firefox, Safari, or Edge.');
-    </script>
 
     <%= yield %>
 
--- a/lib/app/views/layout.erb
+++ b/lib/app/views/layout.erb
@@ -35,7 +35,32 @@
         </div>
       </nav>
     <% end %>
+    <script type="text/javascript">
 
+      /* Cannot put this in app.js because that may not be parseable by older IEs */
+      /* Show alert for IE10 and below */
+
+      function getIEVersion(){
+          var agent = navigator.userAgent;
+          var reg = /MSIE\s?(\d+)(?:\.(\d+))?/i;
+          var matches = agent.match(reg);
+          if (matches != null) {
+              return { major: matches[1], minor: matches[2] };
+          }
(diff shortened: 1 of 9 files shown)


arscan:
I tried it out on a VM with IE10 which caused me to realize why we didn't include the browser alert in `app.js` (because pre-IE11 can't parse it), so I moved it back and tweaked the browser agent string parser so that it behaves appropriately for <IE11 (in Win7), IE11 (in Win8), and IE11 (in Win10).

See screenshot below for the behavior side-by side (alert for <IE11, works in IE11 in both Win8 and Win10).
side-by-side

I think a visual review of the code, and ensuring this doesn't break chome/firefox is an appropriate level of testing -- going through all the work of setting up these legacy IEs again is not necessary.

Jammjammjamm (approved):
Everything seems good in FF/Chrome.

okeefm (approved):
Tested in Chrome on Mac. Everything still works as expected; standalone/EHR launches, single patient, bulk data, and attestation tests all successful.

PR #337, opened 2021-08-06
FI-1258 support newer versions of terminology

okeefm:
# Summary
This PR updates the terminology build process to accept multiple versions of the UMLS zip file, and merges all valid codes from the multiple versions when building the various validators.

## New behavior
* `./bin/run_terminology` now downloads and builds databases for UMLS zip files from 2019, 2020, and 2021
* Building validators through `Inferno::Terminology` can now merge new codes into an existing validator file for Bloom Filters and CSVs
* Rake tasks have been updated to accept version parameters where applicable (defaulting to `2019` for backwards-compatibility)
* UMLS downloads can once again be automated through the provision of a UMLS API key to the Rake task/script

## Code changes
* See above

## Testing guidance
* Testing this one is going to be a bear. The "best" way to test it is to remove `tmp/terminology` and `resources/terminology/validators/bloom/` entirely, then run `./bin/run_terminology.sh` using an API key. This may take 2-3 hours.


bin/run_terminology.sh
```diff
@@ -11,73 +11,34 @@ if [ -f /.dockerenv ]; then
   fi
 fi
 
-temp_folder="tmp/terminology"
-
-if [[ ! -f "$temp_folder/umls.zip" ]]; then
-  echo "The UMLS zip file must be manually downloaded and placed in the tmp/terminology folder."
-  echo "For more info, see the 'terminology support' section of the README."
-  exit 1
+./bin/prepare_terminology.sh 2019
```
Jammjammjamm, line 14: It would be awesome if we could have some sort of config file with the years we wanted so that we wouldn't have to make actual code changes to update these. I guess it would also have to manage AA/AB.

Jammjammjamm (approved):
This works for me, but we should verify that the contents of https://github.com/onc-healthit/inferno-program/blob/main/lib/app/views/server_state.erb are correct

Issue #340, opened 2021-08-09
Race Condition in reference resolution tests (FI-1305)


Jammjammjamm:
There is a race condition in `FHIR::Reference#read`, which can cause the reference resolution tests to fail.

Zulip thread: https://chat.fhir.org/#narrow/stream/179309-inferno/topic/Reference.20server.20tests.20fail.20randomly

PR #342, opened 2021-08-12
FI-1305: Fix race condition

Jammjammjamm:
Updates `fhir_client` to fix race condition in reference resolution tests. This fix broke the argonaut unit tests, so I removed those since they're not needed for program.

Addresses #340.


yunwwang (approved):

radamson (approved):
Looks good. Curious as to what you think about completely stripping the argounaut Inferno tests since their only purpose was to be available for the unit tests (which are now deleted).

Issue #312, opened 2021-07-01
USCMR-13 (MedReq) attempting only intent=plan&status=active,on-hold and failing despite other resources being testable for composite-or (FI-1350)


emily-alt:
Thank you for reporting a possible bug in Inferno! Please fill in as much of the template below as you can.

**Subject of the issue**
"USCMR-13: The server returns results when parameters use composite-or" attempts only 
"GET   200   https://fhir-myrecord-sc.stagingcerner.com/r4/3f2aca24-87f3-4eac-a6d7-1f75247e6b43/MedicationRequest?intent=plan&patient=A879904FD2FE4B2D90C89FDA84E1285F&status=active,on-hold" which returns no on-hold results as expected. 

Patient has resource with status=on-hold but for intent=order. No other combinations are attempted and the test is failed. 

![image](https://user-images.githubusercontent.com/75340305/124167526-1ba72d80-da72-11eb-8738-7625cf31b071.png)

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 1.6.1
* Which browser and version(s) is the bug present on?: Chrome

**Steps to reproduce**
Patient has 5 total MedReq resources available: 
- 1 with intent=order and 4 with intent=plan
- 2 status=active, 1 status=on-hold, 2 status=stopped
Summarized as: 
ACTIVE
A879904FD2FE4B2D90C89FDA84E1285F.17863 Aspirin 81 MG Delayed intent = order
A879904FD2FE4B2D90C89FDA84E1285F.17864 celecoxib 200 MG Oral intent= plan

ON HOLD
A879904FD2FE4B2D90C89FDA84E1285F.17867 Aluminum Hydroxide 80 MG/ML /  Magnesium Hydroxide 80 MG/ML / Simethicone 8 MG/ML Oral Suspension intent = order

STOPPED 
A879904FD2FE4B2D90C89FDA84E1285F.17865 ondansetron HCl intent=order
A879904FD2FE4B2D90C89FDA84E1285F.17866 alum-mag hydroxide-simeth intent=order

### JSON response for intent=order
  GET   200   https://fhir-myrecord-sc.stagingcerner.com/r4/3f2aca24-87f3-4eac-a6d7-1f75247e6b43/MedicationRequest?intent=order&patient=A879904FD2FE4B2D90C89FDA84E1285F
{
  "resourceType": "Bundle",
  "id": "43a7e2b6-c25c-4a81-bff3-93cda93398be",
[...]

Issue #310, opened 2021-06-25
DocumentReference search by Patient, Category and Date fails (FI-1268)


lakshmib08:
Thank you for reporting a possible bug in Inferno! Please fill in as much of the template below as you can.

**DocumentReference search by Patient, Category and Date fails**
Inferno is running a search with a combination of patient, category and date that does not exist in our system.

**Your environment**
Program
* Version of inferno:
inferno.healthit.gov

**Steps to reproduce**
The patient search in step 1 returns 3 records -
The 1st record has - 
**_"category": [
{
"text": "Untitled"
}
],
"subject": {
"reference": "Patient/332769c3-f9e1-4b3d-b1e8-a7efa5ffd25d"
},
"date": "2021-02-04T00:10:07.49+00:00",_**

and the 3rd record has
**_"category": [
{
"coding": [
{
"system": "http://hl7.org/fhir/us/core/CodeSystem/us-core-documentreference-category",
"code": "clinical-note",
"display": "Clinical Note"
}
],
"text": "Clinical Note"
}
],
"subject": {
"reference": "Patient/332769c3-f9e1-4b3d-b1e8-a7efa5ffd25d"
},
"date": "2021-03-02T04:34:05.313+00:00",_**

Inferno searches on -
**_"category=clinical-note&date=2021-02-04T00:10:07.49%2B00:00&patient=332769c3-f9e1-4b3d-b1e8-a7efa5ffd25d"
This search is not valid because no such record exists._**

Thanks.

yunwwang:
It has been identified as a bug in Inferno system. Inferno takes search parameter value from the results of the first search. When there are more than one search parameter, other than patient, the search value may be drawn from different resource instances if one instance does not have all matched elements. Will fix in the next release.

yunwwang:
Working ticket: FI-1268

lakshmib08:
@yunwwang - When can I test this? Thanks.

yunwwang:
Hello @lakshmib08 
If you use Inferno source codes, you can pull the latest from main branch and run tests locally.
If you use Inferno docker image or inferno.healthit.gov website, you have to wait for the v1.7.0 release. We are currently working through the release process. I hope it could be done by next weekend.

PR #343, opened 2021-08-16
FI-1268 Update uscore_generator to prevent get searching value from different…

yunwwang:
… resource when there are more than 2 search parameters

# Summary
When there are more than two search parameters, in some rare case, Inferno may take the search value from different resource under the same patient.

This fix is to force the get_value_for_search_param() to take search parameter values from the same resource 

This PR address Issue #310 and #312

## Code changes
Add an inner loop inside `patient_ids.each do |patient|` that looping through each resource the patient has to find search parameter values

Example:
```
        patient_ids.each do |patient|
          next unless @procedure_ary[patient].present?

          Array.wrap(@procedure_ary[patient]).each do |procedure|
            search_params = {
              'patient': patient,
              'code': get_value_for_search_param(resolve_element_from_path(procedure, 'code') { |el| get_value_for_search_param(el).present? }),
              'date': get_value_for_search_param(resolve_element_from_path(procedure, 'performed') { |el| get_value_for_search_param(el).present? })
            }
...
```

## Testing guidance
Create a new patient in reference server
Create two DocumentReference resource where the first DocumentReference resource has `DocumentReference.category` and `DocumentReference.date` as
```
...
    "category": [
        {
            "text": "Unknown"
        }
    ],
    "date": "2019-09-09",
...
```
and the second DocumentReference resource has
```
    "category": [
        {
            "coding": [
                {
                    "system": "http://hl7.org/fhir/us/core/CodeSystem/us-core-documentreference-category",
                    "code": "clinical-note",
                    "display": "Clinical Note"
                }
            ],
            "text": "Clinical Note"
        }
    ],
    "date": "2020-09-09",
```
Run the "Single Patient" test 
Verify that all tests passed (except secured by TLS test if the reference server is running on http)
Verify that DocumentReference test USCDR-04 has the correct search parameter values like `DocumentReference?category=clinical-note&date=2020-09-09&patient=[patient id]`

### DIFF
--- a/lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
+++ b/lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
@@ -286,33 +286,39 @@ module Inferno
         patient_ids.each do |patient|
           next unless @observation_ary[patient].present?
 
-          search_params = {
-            'patient': patient,
-            'category': get_value_for_search_param(resolve_element_from_path(@observation_ary[patient], 'category') { |el| get_value_for_search_param(el).present? }),
-            'date': get_value_for_search_param(resolve_element_from_path(@observation_ary[patient], 'effective') { |el| get_value_for_search_param(el).present? })
-          }
+          observation_ary = Array.wrap(@observation_ary[patient])
 
-          next if search_params.any? { |_param, value| value.nil? }
+          observation_ary.each do |observation|
+            search_params = {
+              'patient': patient,
+              'category': get_value_for_search_param(resolve_element_from_path(observation, 'category') { |el| get_value_for_search_param(el).present? }),
+              'date': get_value_for_search_param(resolve_element_from_path(observation, 'effective') { |el| get_value_for_search_param(el).present? })
+            }
 
-          resolved_one = true
+            next if search_params.any? { |_param, value| value.nil? }
 
-          reply = get_resource_by_params(versioned_resource_class('Observation'), search_params)
+            resolved_one = true
 
-          reply = perform_search_with_status(reply, search_params, search_method: :get) if reply.code == 400
+            reply = get_resource_by_params(versioned_resource_class('Observation'), search_params)
 
-          validate_search_reply(versioned_resource_class('Observation'), reply, search_params)
+            reply = perform_search_with_status(reply, search_params, search_method: :get) if reply.code == 400
 
-          ['gt', 'ge', 'lt', 'le'].each do |comparator|
-            comparator_val = date_comparator_value(comparator, resolve_element_from_path(@observation_ary[patient], 'effective') { |el| get_value_for_search_param(el).present? })
-            comparator_search_params = search_params.merge('date': comparator_val)
-            reply = get_resource_by_params(versioned_resource_class('Observation'), comparator_search_params)
-            validate_search_reply(versioned_resource_class('Observation'), reply, comparator_search_params)
+            validate_search_reply(versioned_resource_class('Observation'), reply, search_params)
+
+            ['gt', 'ge', 'lt', 'le'].each do |comparator|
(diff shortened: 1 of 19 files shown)


lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
```diff
@@ -286,33 +286,39 @@ def patient_ids
...
-          next if search_params.any? { |_param, value| value.nil? }
+          observation_ary.each do |observation|
+            search_params = {
+              'patient': patient,
+              'category': get_value_for_search_param(resolve_element_from_path(observation, 'category') { |el| get_value_for_search_param(el).present? }),
+              'date': get_value_for_search_param(resolve_element_from_path(observation, 'effective') { |el| get_value_for_search_param(el).present? })
+            }
 
-          resolved_one = true
+            next if search_params.any? { |_param, value| value.nil? }
```
Jammjammjamm, line 298: Should this be a `.present?` check to match the checks right above it.

lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
```diff
@@ -286,33 +286,39 @@ def patient_ids
         patient_ids.each do |patient|
           next unless @observation_ary[patient].present?
 
-          search_params = {
-            'patient': patient,
-            'category': get_value_for_search_param(resolve_element_from_path(@observation_ary[patient], 'category') { |el| get_value_for_search_param(el).present? }),
-            'date': get_value_for_search_param(resolve_element_from_path(@observation_ary[patient], 'effective') { |el| get_value_for_search_param(el).present? })
-          }
+          observation_ary = Array.wrap(@observation_ary[patient])
```
Jammjammjamm, line 289: I think having `observation_ary` and `@observation_ary` will be pretty confusing. Could we use `observations` or `patient_observations` instead? I'm pretty sure I checked a while ago and just adding an `s` works to pluralize all the resources we care about.

Jammjammjamm (commented):
I only looked at the generated code in the sequence I commented on. Do I need to look at others in order to see the full picture for the generated code?

lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
yunwwang: This statement means that if any search_params has nil value, skips the rest and goes to the next observation resource.

Jammjammjamm: I was concerned because `.nil?` isn't the opposite of `.present?` which is used above. But I guess for this check it's ok because any blank value would end up getting turned into nil.

yunwwang: Good question. The statement was used before so I didn't change it to avoid any unexpected side effects. 
Logically, `next if search_params.any? { |_param, value| value.nil? }` is the same as `next unless search_params.any? { |_param, value| value.present? }` other than when value==''. Hard decision.

Jammjammjamm: But if `value.present?` never returns true, then I think `resolve_element_from_path` will return `nil`, so it should be fine.

lib/modules/uscore_v3.1.1/bodyheight_sequence.rb
yunwwang: Combined this line with next line to
```
Array.wrap(@observation_ary[patient]).each do |observation|
```

Jammjammjamm (approved):

yunwwang:
All search changes generated by the same generator function. CareTam and MedicationRequest have composite-or searches which are also modified by a different generator function but under the similar pattern

czh-orz (approved):

PR #347, opened 2021-08-19
FI-1271 Update links to US Core 3.1.1. to include "STU3.1.1"

yunwwang:
# Summary
US Core points the default url `http://hl7.org/fhir/us/core/... points` to latest published version
After UScore 4.0.0 release, Inferno need to change its quoted url to v3.1.1 profiles

## New behavior
Add `/STU3.1.1` to browserable url link while keep the profile uri unchanged.

## Code changes
Update medata_extractor to build a "versioned_profile" by inserting `/STU3.1.1` in the middle of profile url
Update uscore_generator to generator test detail and link section to have the correct browserable url

## Testing guidance
Open Single Patient and Multi Patient test and verify that links in the test are redirect to US Core v3.1.1 profiles
Only urls in the text area (links, details) are changed. This PR does not affect any testing steps.

### DIFF
--- a/lib/modules/uscore_v3.1.1/access_verify_restricted_sequence.rb
+++ b/lib/modules/uscore_v3.1.1/access_verify_restricted_sequence.rb
@@ -96,7 +96,7 @@ module Inferno
           id '01'
           name 'Scope granted is limited to those chosen by user during authorization.'
 
-          link 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient'
+          link 'http://hl7.org/fhir/us/core/STU3.1.1/StructureDefinition/us-core-patient'
           description %(
 
             This test confirms that the scopes granted during authorization match those that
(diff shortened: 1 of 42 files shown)


lib/modules/uscore_v3.1.1/access_verify_restricted_sequence.rb
```diff
@@ -96,7 +96,7 @@ def scope_granting_access(resource, scopes)
           id '01'
           name 'Scope granted is limited to those chosen by user during authorization.'
 
-          link 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient'
+          link 'http://hl7.org/fhir/us/core/STU3.1.1/StructureDefinition/us-core-patient'
```
Jammjammjamm, line 99: This doesn't seem like the correct link for this test.

Jammjammjamm (changes requested):
The python submodule is here, too.

PR #351, opened 2021-08-25
FI-1315 Add "usg: sig" to bulk data's public key

yunwwang:
# Summary
This PR addresses GitHub Issue #344 

## New behavior
The JWKS  at [base]/inferno/.well-known/jwks.json contains both key_ops and use in public key.

Example:

```
            "use": "sig",
            "key_ops": [
                "verify"
            ], 
```
## Code changes
Updated public keys in config.yml

## Testing guidance
Verify [base]/inferno/.well-known/jwks.json contains `"use": "sig",`
Verify Bulk Data Authorization sequence runs without error


sagarshah1983:
Is this the approach inferno planning to make? Or are there any other considerations? Are we waiting on anything else on this PR?
Please advise so we can plan testing on this accordingly.

czh-orz (approved):

PR #352, opened 2021-08-27
FI-1316 Call bundle.next_bundle to get next page of search result

yunwwang:
# Summary
USCMR-14 failed because the resources from the 2nd page of search result don't have fhir client reference.
The root cause is on the SequenceBase.fetch_all_bundled_resources() function which does not use parse_reply for the response from "next" link.

## New behavior
Use bundle.next_bundle to get next page of search result.
bundle.next_bundle return nil if there is no next page or unable to read next page.

Remove one unit test on status 404 since bundle.next_bundle does not throw exception on status 404. Instead it returns nil.

## Testing guidance
Run Single Patient test on Reference Server to verify that USCMR-14 is passed.

### DIFF
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -886,6 +886,7 @@ module Inferno
           reply = @client.raw_read_url(next_bundle_link)
           error_message = "Could not resolve next bundle. #{next_bundle_link}"
           assert_response_ok(reply, error_message)
+          assert_bundle_response(reply)
 
           bundle = @client.parse_reply(FHIR::Bundle, @client.default_format, reply)
 
--- a/lib/app/sequence_base.rb
+++ b/lib/app/sequence_base.rb
@@ -878,17 +878,8 @@ module Inferno
         bundle = reply.resource
         until bundle.nil? || page_count == 20
           resources += bundle&.entry&.map { |entry| entry&.resource }
-          next_bundle_link = bundle&.link&.find { |link| link.relation == 'next' }&.url
           reply_handler&.call(reply)
-          break if next_bundle_link.blank?
-
-          reply = @client.raw_read_url(next_bundle_link)
-          error_message = "Could not resolve next bundle. #{next_bundle_link}"
-          assert_response_ok(reply, error_message)
-          assert_valid_json(reply.body, error_message)
-
-          bundle = FHIR.from_contents(reply.body)
-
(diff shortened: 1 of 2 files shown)


test/unit/sequence_base_test.rb
```diff
@@ -333,15 +333,6 @@ class SequenceBaseTest < MiniTest::Test
       assert all_resources.map(&:id) == ['1', '2']
     end
 
-    it 'fails on 404' do
```
Jammjammjamm, line 336: This means we no longer fail a server if the next link in a Bundle is broken.

hershilpatel (approved):
Tested and works

test/unit/sequence_base_test.rb
yunwwang: Yes. the next_bundle returns nil if the next_link is empty or the server returns 4xx

Jammjammjamm: I think that is a problem.

yunwwang: Restored the unit test and changed the fetch_all_bundled_resources() to handle that

lib/app/sequence_base.rb
```diff
@@ -880,14 +880,14 @@ def fetch_all_bundled_resources(reply, reply_handler = nil)
           resources += bundle&.entry&.map { |entry| entry&.resource }
           next_bundle_link = bundle&.link&.find { |link| link.relation == 'next' }&.url
           reply_handler&.call(reply)
+
           break if next_bundle_link.blank?
 
           reply = @client.raw_read_url(next_bundle_link)
           error_message = "Could not resolve next bundle. #{next_bundle_link}"
           assert_response_ok(reply, error_message)
-          assert_valid_json(reply.body, error_message)
```
Jammjammjamm, line 888: Why don't we need this assertion anymore?

yunwwang: If this body is invalid, @client.parse_reply would return nil
https://github.com/fhir-crucible/fhir_client/blob/42c675db614dd20812f617181b119af4a7d1b000/lib/fhir_client/client.rb#L315

Jammjammjamm: So now a server would pass if the next link in a bundle didn't return a body?

Jammjammjamm (approved):

PR #354, opened 2021-09-02
FI-1304 Adding Observation.component MS check for BP profile. Remove dataAbsentReason MS check from all vital sign profiles

czh-orz:
# Summary
This PR fixes GitHub issue #385, #364

Inferno decides to remove MustSupport check for Observation.dataAbsentResent and relies on invariant vs-2 to validate that "If there is no component or hasMember element then either a value[x] or a data absent reason must be present."

## New behavior
Add Observation.component MS check for observation-bp profile
Remove Observation.dataAbsentReason MS check from vital sign profile

## Code changes
add logic to remove dataAbsentReason to metadata_extractor

## Testing guidance
Check if tests still pass

### DIFF
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -274,10 +274,6 @@ module Inferno
       def add_must_support_elements(profile_definition, sequence)
         profile_elements = profile_definition['snapshot']['element']
         profile_elements.select { |el| el['mustSupport'] }.each do |element|
-          # not including components in vital-sign profiles because they don't make sense outside of BP
-          next if profile_definition['name'] == 'observation-bp' && element['path'].include?('Observation.value[x]')
-          next if profile_definition['name'].include?('Pediatric') && element['path'] == 'Observation.dataAbsentReason'
-
           if element['path'].end_with? 'extension'
             sequence[:must_supports][:extensions] <<
               {
@@ -527,12 +523,6 @@ module Inferno
         medication_request_sequence = metadata[:sequences].find { |sequence| sequence[:resource] == 'MedicationRequest' }
         set_first_search(medication_request_sequence, ['patient', 'intent'])
 
-        pulse_ox_sequence = metadata[:sequences].find { |sequence| sequence[:profile] == 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-pulse-oximetry' }
-        pulse_ox_sequence[:must_supports][:elements].delete_if do |element|
-          path = element[:path]
-          (path.start_with? 'component.value') && (!path.include? '[x]') && (path != 'component.valueQuantity')
-        end
-
         # remove carrierAIDC and carrierHRF from implantable device must supports
         implantable_device_sequence = metadata[:sequences].find { |sequence| sequence[:profile] == 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-implantable-device' }
         implantable_device_sequence[:must_supports][:elements].delete_if do |element|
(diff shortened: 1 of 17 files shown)


Jammjammjamm:
What about the pulse ox profile?

czh-orz:
> What about the pulse ox profile?

Pulse ox already had the component must support elements.

Jammjammjamm (approved):

arscan:
Is it too much work to have this be handled in the `add_special_cases` section of the generator, instead of having these special cases scattered throughout?  It is much easier for others to understand how we've deviated from the machine-readable constraints if we make all the tweaks at the end (I know it is more code, but I think clarity is more important here).

This may not be practical for this type of constraint though -- I know our must support checking stuff is a bit separate.  Just checking.

czh-orz:
I think it makes sense to move this to the special cases section, but it will take a bit of a shuffle.

czh-orz:
> Is it too much work to have this be handled in the `add_special_cases` section of the generator, instead of having these special cases scattered throughout? It is much easier for others to understand how we've deviated from the machine-readable constraints if we make all the tweaks as the very last step (I know it is more code, but I think clarity is more important here).
> 
> This may not be practical for this type of constraint though -- I know our must support checking stuff is a bit separate. Just checking.

I moved the logic to the special cases section

arscan:
> What about the pulse ox profile?

I was trying to understand why this didn't affect pulse ox, and its because pulse ox doesn't build directly off the vital signs, it builds off another profile that builds off vital signs, so it doesn't get hit by that filter throwing out components.

However, in looking at the Pulse Ox profile, why do we require showing both component.value and value?  Is that right?  I thought it was generally one or the other?  And if its wrong, why do we pass our reference server?

https://github.com/onc-healthit/inferno-program/blob/c7b623369bcbcba4d691ba2b374d9f1258f2bd4e/lib/modules/uscore_v3.1.1/us_core_pulse_oximetry_sequence.rb#L654-L686

lib/modules/uscore_v3.1.1/profile_definitions/bp_definitions.rb
```diff
@@ -53,6 +87,32 @@ class USCore311BpSequenceDefinitions
           }.freeze,
           {
             path: 'dataAbsentReason'
```
yunwwang, line 89: Since we exclude Obervation.value[x], should we also exclude Observation.databasentReason?

czh-orz:
> > What about the pulse ox profile?
> 
> I was trying to understand why this didn't affect pulse ox, and its because pulse ox doesn't build directly off the vital signs, it builds off another profile that builds off vital signs, so it doesn't get hit by that filter throwing out components.
> 
> However, in looking at the Pulse Ox profile, why do we require showing both component.value and value? Is that right? I thought it was generally one or the other? And if its wrong, why do we pass our reference server?
> 
> https://github.com/onc-healthit/inferno-program/blob/c7b623369bcbcba4d691ba2b374d9f1258f2bd4e/lib/modules/uscore_v3.1.1/us_core_pulse_oximetry_sequence.rb#L654-L686

Since component.value and value are both must support, doesn't that mean we have to see both? Our server returns a pulse ox observation with both a component.valueQuantity and a valueQuantity. 
So for pulse ox, are we not supposed to test the component must support? I'm not sure what the intent here is.

arscan:
> Since component.value and value are both must support, doesn't that mean we have to see both? Our server returns a pulse ox observation with both a component.valueQuantity and a valueQuantity.
So for pulse ox, are we not supposed to test the component must support? I'm not sure what the intent here is.

The challenge is that is appears that for many of the Observation profiles, inclusion of must support for both `component.value[x]` and `value[x]` is clearly incorrect, but it occurs because of technicalities of how profiles inherit must support flags from parent profiles (e.g. vital signs, which flags both, because some vital signs need component, some need non-component... and maybe some require both?).  I do not know for pulse ox if both make sense or not to include.  Our test data does, which seems to indicate it makes sense, but I want to make sure that isn't an error too...

generator/uscore/metadata_extractor.rb
```diff
@@ -558,10 +548,36 @@ def add_special_cases(metadata)
...
+        bp_sequence[:must_supports][:elements].delete_if do |element|
+          element[:path].start_with?('value[x]')
+        end
+        bp_sequence[:must_supports][:slices].delete_if do |slice|
+          slice[:path].start_with?('value[x]')
+        end
+
+        # ONC clarified that health IT developers that always provide HL7 FHIR "observation" values
+        # are not required to demonstrate Health IT Module support for "dataAbsentReason" elements.
+        # Remove MS check for dataAbsentReason and component.dataAbsentReason from vital sign profiles
```
Jammjammjamm, line 568: This is only for vital signs, not for all observations?

yunwwang: Good catch! My first impression is that this is only for vital sign profiles. After reading CCG again, I think it means all Observation profiles.

Jammjammjamm (approved):

generator/uscore/metadata_extractor.rb
```diff
@@ -546,6 +535,45 @@ def add_special_cases(metadata)
         document_reference_sequence[:must_supports][:elements].delete_if do |element|
           ['content.attachment.data', 'content.attachment.url'].include? element[:path]
         end
+
+        # exclude component from vital sign profiles except observation-bp and observation-pulse-ox
+        # observation-bp is excluded by profile_definition['differential']['element'].any? { |el| el['path'] == 'Observation.component' }
+        # observation-plux-ox is excluded by profile_definition['baseDefinition'] == 'http://hl7.org/fhir/StructureDefinition/vitalsigns'
```
arscan, line 541:
```suggestion
        # observation-pulse-ox is excluded by profile_definition['baseDefinition'] == 'http://hl7.org/fhir/StructureDefinition/vitalsigns'
```

arscan (approved):

yunwwang:
Migrated to Inferno v2.0 uscore-test-kit: https://github.com/inferno-framework/us-core-test-kit/pull/10

Issue #355, opened 2021-09-02
Must-support tests that should be failing are passing incorrectly (FI-1321)


cooperthompson:
**Subject of the issue**
Our FHIR server mysteriously starting passing several Inferno tests that we did not expect it to pass (because we have some dev in flight or are waiting on some Inferno fixes documented in other github issues).

These are some example tests that are passing incorrectly.  It appears these are all related to must-support checks.

USCG-08
USCDR-13
USCID-08
USCP-13

Here is an example test session:
https://inferno.healthit.gov/inferno/e7X7MhhmJlv/test_sets/test_procedure/

**Your environment**
* Edition of inferno (Community or Program):  Program
* Version of inferno: 1.6.2
* Which browser and version(s) is the bug present on?: Chrome

**Steps to reproduce**
Tell us how to reproduce this issue.

Include:

* The selected test suite
* The URL of the FHIR server being tested:  Inferno Logo Small https://hit-cert.epic.com:444/FHIRARR98/api/FHIR/R4
* The Test ID, if it's a problem with a specific test. Listed above/
* Any required configuration options (client ID, client secret, etc.)  
* Anything else needed to make the issue occur.

**Expected behavior**
Tell us what should happen.
The tests listed above should fail, based on the returned FHIR content.

**Actual behavior**
Tell us what happens instead. Include screenshots if possible.
The tests pass unexpectedly.

cooperthompson:
Created per @Jammjammjamm.  See https://chat.fhir.org/#narrow/stream/179309-inferno/topic/Intermittently.20passing.20Inferno.20when.20we.20shouldn't.20be

Jammjammjamm:
I've identified the cause of this issue and am working on a fix.

PR #357, opened 2021-09-09
FI-1321: Fix Must Support tests incorrectly passing

Jammjammjamm:
# Summary
Addresses #355. The bulk group validation sequence was modifying the US Core profile definitions. Any MS field found in that sequence would not be checked in any future test runs.

## Code changes

- US Core profile definitions are now deep frozen, so any attempt to modify them will raise an error
- The bulk group validation sequence now deep dups the profile definitions so that the originals will not be modified

## Testing guidance

- Run the single patient API tests with patient 85. USCDR-13 should skip.
- Run the bulk tests
- Run the single patient API tests with patient 85 again. USCDR-13 should still skip. If you follow these three steps on main, it will pass.

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -133,7 +133,7 @@ module Inferno
             }
             delayed_profile_intersection
           end
-          sequence[:delayed_references_constant] = "DELAYED_REFERENCES = #{structure_to_string(delayed_sequence_references)}.freeze"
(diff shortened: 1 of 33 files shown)


arscan (approved):
I was able to reproduce the issue on `main` and confirm that it is gone in this branch.  The code changes look reasonable to me.  I think we need to make sure this makes it into the next release.

yunwwang (approved):

PR #358, opened 2021-09-10
Update fhir client in pg gemfile

Jammjammjamm:
Updates the Gemfile.postgres and Gemfile.postgres.lock, which should have been done in #342.

To test in the most basic way, run:
```
docker-compose up --build
```
and Inferno should run.

To test in a more thorough manner, undo the change introduced in https://github.com/onc-healthit/inferno-program/pull/352, then launch with ``docker-compose up --build``, then run the single patient api tests.  You should see the following for medication request:

Screen Shot 2021-09-10 at 10 39 02 AM

Then, when you change that line of code back (undo your change), that test should now pass.

I do not think we need to update the changelog with this update, since this should have just been included in #342


arscan (approved):
Fixes oversight from update in https://github.com/onc-healthit/inferno-program/pull/342

arscan:
FYI @yunwwang I was able to recreate the MedicationRequest issue you were seeing (and that we have subsequently fixed) using docker after this update, so this resolves that outstanding question about why you were the only one seeing that issue.

czh-orz (approved):

radamson (approved):

Issue #359, opened 2021-09-15
Bulk data export status check (BDE-07) timeout should be configurable (FI-1327)


yunwwang:
Current BDE-07 timeout after 180 seconds. The timeout is calculated as
```
CurrentTime - StartTime + WaitTime
```
where the WaitTime is the value from `retry_after` header (if exist) or the exponential backoff from previous WaitTime

Some server may need more time than 180 seconds to get export ready. 

Inferno should allow tester to manually change the timeout threshold at run time, (ex via UI component)

medhost-gkersting:
I was also curious about the ETA on this fix.

yunwwang:
@jhrpcek @medhost-gkersting  Inferno v1.8.0 is released. Could you verify that this issue has been addressed in this release?

jhrpcek:
We don’t see any new failures introduced and the configurable time-out for multi-patient seems to work.

PR #382, opened 2021-10-20
FI-1327 Add UI input field for bulk_timeout

yunwwang:
# Summary
This PR address GitHub issue #359 

We added a UI input field for Bulk Data export status check timeout value

## New behavior
Tester can specify maximum timeout value for Bulk Data export status check.
If tester input value less than 1 or NAN, Inferno uses default 180 seconds.

## Testing guidance
When starting a Multi-Patient API test, the "Export times out after" shows at the bottom of popup dialog with default value 180.

### DIFF
--- a/lib/modules/onc_program/bulk_data_export_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_export_sequence.rb
@@ -91,7 +91,7 @@ module Inferno
 
       def check_export_status(url = @content_location, timeout: 180)
         skip 'Server response did not have Content-Location in header' unless url.present?
-        
+
         timeout = @instance.bulk_timeout if @instance.bulk_timeout.positive?
         reply = export_status_check(url, timeout)
 
--- a/lib/modules/onc_program/bulk_data_export_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_export_sequence.rb
@@ -15,7 +15,7 @@ module Inferno
 
       test_id_prefix 'BDE'
 
-      requires :bulk_url, :bulk_access_token, :bulk_lines_to_validate
+      requires :bulk_url, :bulk_access_token, :bulk_lines_to_validate, :bulk_timeout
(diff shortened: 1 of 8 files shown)


arscan:
@Jammjammjamm does this require any special instructions in the release notes (it has a schema change), or will it auto-migrate based on however we are launching inferno?

lib/modules/onc_program/bulk_data_export_sequence.rb
```diff
@@ -91,6 +91,8 @@ def assert_kick_off_error_response(reply)
 
       def check_export_status(url = @content_location, timeout: 180)
         skip 'Server response did not have Content-Location in header' unless url.present?
+
+        timeout = @instance.bulk_timeout if @instance.bulk_timeout.positive?
```
arscan, line 95: Should we put a max on this?  Or are we ok with letting tests essentially hang until next reboot if too big a number was put here?

arscan (approved):
Looks good.  I'm not sure if my comment about capping the timeout to a max value would realistically make a difference, as there are other opportunities for malicious testers to cause very long-running tests.  So I leave it up to you if you want to implement it or not.  Good to think defensively in general though.

jhrpcek:
Any ETA on when the new release will be available with this fix?   Prior messages said this past weekend it should be out??

yunwwang:
Sorry, we are delayed a little bit by testing of new terminology deployment.

lib/modules/onc_program/bulk_data_export_sequence.rb
yunwwang: I will create another ticket to address this because there's similar complexity that how long is long enough and should that value be configurable also.

yunwwang:
Migrated to g10-certification-test-kit: https://github.com/inferno-framework/g10-certification-test-kit/pull/19

Issue #390, opened 2021-11-05
BDE-02: Incorrectly fails this test case (Bulk Data Server declares support for Group export operation in CapabilityStatement) (FI-1366)


sagarsarvankar:
Thank you for reporting a possible bug in Inferno! Please fill in as much of the template below as you can.

**BDE-02 fails even when we provide correct URL+"|"+version for operation.definition **

The spec for operation.definition (https://www.hl7.org/fhir/operationdefinition.html) says that this can be a target of a canonical reference. Our FHIR server has a operation.definition value of "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export|1.0.1", which is acceptable going by the spec.

{
"type": "Group",
"profile": "http://hl7.org/fhir/profiles/Group",
"operation": [{
"name": "export",
"definition": "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export|1.0.1"
}]
}

When we run Inferno, this test case fails, it expects the value to be exact "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export"

Could you please address this issue?

**Your environment**
* Edition of inferno (Community or Program): https://inferno.healthit.gov/inferno/
* Version of inferno: 1.7.0
* Which browser and version(s) is the bug present on?: Chrome.

**Steps to reproduce**
Tell us how to reproduce this issue.

Include:

* The selected test suite
* The URL of the FHIR server being tested      https://fhir.secureit.co.in:9443/fhir-server/api/v4   We keep this server active only when we need it. For today we can keep this on, after that whenever you need it active, please let us know and we will do so.
* The Test ID, if it's a problem with a specific test.    BDE-02

**Expected behavior**
This should accept definition value "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export|1.0.1" 

**Actual behavior**
Inferno is marking this test case as a fail, and checking against the exact value "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export"

yunwwang:
CapabilityStatement.rest.resource.operation.definition is a type of `canonical(OperationDefinition)`. Inferno should support canonical url with `|version` postfixed.

yunwwang:
@sagarsarvankar This issue has been fixed in Inferno Program v1.8.2. Would you mind to check if this fix addressed your concern? Thank you.

sagarsarvankar:
Thank you @yunwwang . We will test the same and let you know what happens soon.

[...]

PR #400, opened 2021-11-15
FI-1366 BED-02 validate canonical url with version 

yunwwang:
# Summary
This is a fix of #390 
FHIR canonical URL allows version number to be attached at the end of url. BED-02 validates that CapabilityStatement has an operation defined by canonical url "http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export"

## New behavior
Using Regular Expression /^http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export(\|\S*)?/ to validate canonical url

## Testing guidance
A new unit test is added to verify that BED-02 passed url with version

### DIFF
--- a/lib/modules/onc_program/bulk_data_export_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_export_sequence.rb
@@ -47,7 +47,6 @@ module Inferno
         conformance = versioned_resource_class.from_contents(reply.body)
         assert conformance.present?, 'Cannot read server CapabilityStatement.'
 
-        # assert operation.present?, 'Server CapabilityStatement did not declare support for export operation in Group resource.'
         has_export = conformance.rest&.any? do |rest|
           rest.resource&.any? do |resource|
             resource.type == 'Group' &&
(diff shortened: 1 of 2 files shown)


lib/modules/onc_program/bulk_data_export_sequence.rb
```diff
@@ -47,18 +47,16 @@ def check_capability_statement
...
-        operation = nil
-
-        conformance.rest&.each do |rest|
-          group = rest.resource&.find { |r| r.type == 'Group' && r.respond_to?(:operation) }
-
-          next if group.nil?
-
-          operation = group.operation&.find { |op| op.definition == 'http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export' }
-          break if operation.present?
+        # assert operation.present?, 'Server CapabilityStatement did not declare support for export operation in Group resource.'
```
Jammjammjamm, line 50: Don't need this.

Jammjammjamm (approved):

lib/modules/onc_program/test/bulk_patient_export_sequence_test.rb
```diff
@@ -56,10 +56,10 @@
     before do
```
arscan, line 56: As an aside, every time i see `before` I have to double check if it is before each example or just once for all examples, because it is important in this case because we are mutating `@conformance`.  I think it would be a good idea to use the explicit `before(:each)` form in general, but i know that's not what we do right now.

arscan (approved):

yunwwang:
g10-certification-test-kit PR: https://github.com/inferno-framework/g10-certification-test-kit/pull/19

Issue #398, opened 2021-11-12
Bulk Testing Client Does Not Follow Redirects (FI-1371)


bforbis:
I am trying to pass the multi-patient test sequences for **BDGV-06** through **BDGV-23**. Initially our service was using `requiresAccessToken=false` and using signed URLs for data access, however we recently got instructions from ONC that this is not allowed per ONC rulings (see #303 ).

Our next approach for implementing bulk data access validations of our file servers was to:
- Have a simple authorizer service which validates the bearer token for a given file
- If the token is valid, returns a 301 redirect with a signed s3 URL to the file server which contains the content

In practice, we expect that most HTTP clients will be able to follow these 301 redirects, however the inferno client does not.

Would Inferno be willing to make a change to have the client follow redirects?

yunwwang:
This is interesting. What is your interpretation on Bulk Data Export Section [5.4] (http://hl7.org/fhir/uv/bulkdata/STU1.0.1/export/index.html#file-request)?

bforbis:
The spec does not classify whether 3xx are considered an error or a success. However, in my experience many http clients will happily follow redirects and it is not unreasonable to assume clients can handle this. Would this need to be raised to HL7 to decide on?

yunwwang:
We had a brief discussion at [zulip](https://chat.fhir.org/#narrow/stream/179250-bulk-data/topic/301.20redirect). One question raised is about security headers. Does your S3 bucket require additional headers?

bforbis:
No, just the signed url is required.

bforbis:
FYI @yunwwang, when dealing with Amazon S3 in particular, there is a requirement to use only one Authorization mechanism. This means that on handling of the 30[1|2|7] redirect, the client must remove the Authorization header.

More details on that requirement here: https://stackoverflow.com/questions/31514336/only-one-auth-mechanism-allowed-only-the-x-amz-algorithm-query-parameter

yunwwang:
Thank you for the latest comment. Looks like our current redirect solution has a side effect. The team will investigate this issue again.

bforbis:
Hi @yunwwang, what is the status with this ticket? Are we going to proceed with this fix, or do you think we may revisit the conversation with allowing for requiresAccessToken=false so that we can return signed URLs directly in the status endpoint?

yunwwang:
[...]

PR #408, opened 2021-11-22
FI-1371 Bulk export redirect

yunwwang:
# Summary
This is a fix for #398 

When Bulk Data client requests a NDJSON file, Bulk data server MAY respond a HTTP redirect with status code 301, 302, 303, or 307 and redirect link in the header['location'].

## New behavior
Inferno bulk data client follows redirect link in the header['location']

## Testing guidance
A new unit test is created to test 301 redirect.

### DIFF
--- a/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
+++ b/lib/modules/onc_program/bulk_data_group_validation_sequence.rb
@@ -382,6 +382,15 @@ module Inferno
           block_response: response_block
         )
 
+        if ['301', '302', '303', '307'].include?(response.code) && response[:location].present?
+          response = RestClient::Request.execute(
+            method: :get,
+            url: response[:location],
+            headers: headers,
+            block_response: response_block
+          )
+        end
+
         response_for_log = {
           code: response.code,
           headers: response.header,
--- a/lib/modules/onc_program/test/bulk_group_validation_sequence_test.rb
+++ b/lib/modules/onc_program/test/bulk_group_validation_sequence_test.rb
@@ -9,8 +9,6 @@ describe Inferno::Sequence::BulkDataGroupExportValidationSequence do
     @patient_file_location = 'https://www.example.com/patient_export.ndjson'
     @condition_file_location = 'https://www.example.com/condition_export.ndjson'
 
-    @patient_redirect_location = 'https://www.example.com/patient_export_redirect.ndjson'
-
     @patient_export = load_fixture_with_extension('bulk_data_patient.ndjson')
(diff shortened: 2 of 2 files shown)


lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -382,6 +382,15 @@ def streamed_ndjson_get(url, headers)
           block_response: response_block
         )
 
+        if ['301', '302', '303', '307'].include?(response.code) && response[:location].present?
+          response = RestClient::Request.execute(
+            method: :get,
+            url: response[:location],
```
arscan, line 388: HTTP 1.1 allows relative urls in the location header.  So you should check to see if it starts with http, and if not, calculate the relative url based on the current location.

lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -382,6 +382,15 @@ def streamed_ndjson_get(url, headers)
           block_response: response_block
         )
 
+        if ['301', '302', '303', '307'].include?(response.code) && response[:location].present?
```
arscan, line 385: Do we care if they redirect to a redirect?  @Jammjammjamm came up with this pseudocode:

```
response = RestClient::Request.execute(...)

while response_is_a_redirect do
  # should also probably specify a maximum
  # redirect count
  response = RestClient::Request.execute(...)
end
```

yunwwang: set maximum redirect count to 10?

lib/modules/onc_program/bulk_data_group_validation_sequence.rb
```diff
@@ -375,13 +375,30 @@ def streamed_ndjson_get(url, headers)
...
           headers: headers,
           block_response: response_block
         )
 
+        while ['301', '302', '303', '307'].include?(response.code) && response[:location].present? && max_redirect.positive?
+          max_redirect -= 1
+          redirect_url = response[:location]
+
+          # handle relative redirects
+          redirect_url = URI.parse(url).merge(redirect_url).to_s unless redirect_url.start_with?('http')
```
Jammjammjamm, line 392: It's hard to be certain whether this works properly, so you should add a unit test for the relative redirect case.

arscan:
We implemented this in [v2.0 of the g10 certification tests](https://github.com/onc-healthit/onc-certification-g10-test-kit). If you need this functionality, please upgrade to the new version of the tests, and let us know if you are having issues with that process.  Thanks!

Issue #412, opened 2021-11-29
DiagnosticReport search validations only check the first entry in CodeableConcept on responses (FI-1406)


cooperthompson:
**Subject of the issue**
When Inferno does code and category searches on DiagnosticReport (labs), it looks like you are validating that the code/category in the response matches what you searched on.  However, it looks like you are only checking the first entry in the CodeableConcept, rather than any entry.

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 1.8.1
* Which browser and version(s) is the bug present on?:  Chrome (current)

**Steps to reproduce**
Run USCDRLRR-03 and USCDRLRR-04 on a server returning DiagnosticReport resources that have multiple values in the code and category properties, where the searched value is not the first value in the list.

Include:

* The selected test suite:  Single Patient API
* The URL of the FHIR server being tested (internal dev server)
* The Test ID, if it's a problem with a specific test.  USCDRLRR-03 and USCDRLRR-04.  Possibly several others that we haven't run into yet if we are returning the expected code in the first element.
* Any required configuration options (client ID, client secret, etc.)
* Anything else needed to make the issue occur.

**Expected behavior**
Tell us what should happen.
When Inferno does a code or category search, it should check if the searched code is in _any_ of the entries in the returned response, not just the first one.

**Actual behavior**
Tell us what happens instead. Include screenshots if possible.
Inferno looks like it is checking only the first entry in the CodeableConcept when validating that a returned resource is correctly included based on code/category search criteria.

Here is the DiagnosticReport search response bundle that is causing the errors:
[dxrpt.json.txt](https://github.com/onc-healthit/inferno-program/files/7619210/dxrpt.json.txt)

![image](https://user-images.githubusercontent.com/5863379/143895468-cae537a9-d9fa-4767-a482-d0210375486b.png)

cooperthompson:
Also, should the comparison for search value to response content be case insensitive for tokens?  Case sensitivity for tokens is talked about in the [token definition](http://hl7.org/fhir/R4/search.html#token), and it isn't really absolute.  It does refer to [CodeSystem.caseSensitive](https://www.hl7.org/fhir/codesystem-definitions.html#CodeSystem.caseSensitive) which references Postel's Law, which implies that a client should accept codes in any case.

yunwwang:
@cooperthompson  I tested this locally (v1.8.1) without getting any error for either of the following combinations:

```       
  "category": [
    {
      "coding": [
        {
[...]

PR #421, opened 2021-12-16
FI-1406 Allow case-insensitive comparison for token search

yunwwang:
# Summary
This PR fixes GitHub Issue #412 

Inferno allows case-insensitive comparison on token value

## Code changes
Fix code generator for CodeableConcept data types in search_parameter_match_found method

### DIFF
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1450,9 +1450,9 @@ module Inferno
             coding_value = value.split('|').last
             match_found = values_found.any? do |codeable_concept|
               if value.include? '|'
-                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code.casecmp?(coding_value) }
+                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code&.casecmp?(coding_value) }
               else
-                codeable_concept.coding.any? { |coding| coding.code.casecmp?(value) }
+                codeable_concept.coding.any? { |coding| coding.code&.casecmp?(value) }
               end
             end)
         when 'Identifier'
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1450,9 +1450,9 @@ module Inferno
             coding_value = value.split('|').last
             match_found = values_found.any? do |codeable_concept|
               if value.include? '|'
-                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code == coding_value }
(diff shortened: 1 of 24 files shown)


generator/uscore/uscore_generator.rb
```diff
@@ -1450,9 +1450,9 @@ def search_param_match_found_code(type, element)
             coding_value = value.split('|').last
             match_found = values_found.any? do |codeable_concept|
               if value.include? '|'
-                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code == coding_value }
+                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code.casecmp?(coding_value) }
```
Jammjammjamm, line 1453: We need to be nil-safe in these checks: `coding.code&.casecmp?`.

generator/uscore/uscore_generator.rb
```diff
@@ -1450,9 +1450,9 @@ def search_param_match_found_code(type, element)
             coding_value = value.split('|').last
             match_found = values_found.any? do |codeable_concept|
               if value.include? '|'
-                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code == coding_value }
+                codeable_concept.coding.any? { |coding| coding.system == coding_system && coding.code&.casecmp?(coding_value) }
               else
-                codeable_concept.coding.any? { |coding| coding.code == value }
+                codeable_concept.coding.any? { |coding| coding.code&.casecmp?(value) }
```
arscan, line 1455: Looks good.  I was a little concerned about the behavior if `coding.code` is `nil` here, but it works out... if `coding.code` is `nil` then the block will return `nil` which is evaluated as `false`, which is what we want.

arscan: I also check what happens if `value` is nil, and if that is the case, then this will also return nil, which is evaluated as false, which works out.  Bottom line, this is fine.  I just wanted to note that I checked those two things :)

arscan (approved):
Code looks good.  I ran the tests, and no issues came up on our data set.  I was wondering if there were other places where this could theoretically come up -- in looking through tests was wondering if `status` was one of them (code comparison checks are different for those), but in thinking it through I don't think its possible for this to be an issue with `status` searches (required binding, and we generally aren't searching by statuses returned by the server -- we hardcode the required statuses in and search by them).

If there are other classes where this is an issue that we've missed somehow, we could address them in the future.

Jammjammjamm (approved):

yunwwang:
This fix is migrated to Inferno v2 us-core-test-kit. PR: https://github.com/inferno-framework/us-core-test-kit/pull/8

Issue #374, opened 2021-10-05
Inferno does not check if the attachment is downloadable (FI-1373)


yunwwang:
Thank you for reporting a possible bug in Inferno! Please fill in as much of the template below as you can.

**Subject of the issue**
Clinical Notes Guidance [requires](http://hl7.org/fhir/us/core/STU3.1.1/clinical-notes-guidance.html#fhir-resources-to-exchange-clinical-notes) that
```quote
Argonaut Clinical Note Server SHALL expose these reports through both 
DiagnosticReport and DocumentReference by representing the same attachment url 
using the corresponding elements listed below:

    DocumentReference.content.attachment.url
    DiagnosticReport.presentedForm.url

```

Currently, Inferno compares url from both DocumentReference and DiagnosticReport but not checking if those urls are downloadable.

**Your environment**
* Edition of inferno (Community or Program): Program
* Version of inferno: 17.0

**Expected behavior**
Inferno should try to download every URL to verify that those urls are real and the attachment is downloadable.

yunwwang:
When download attachment, we should consider adding patient/Binary.read to the token request so that Inferno is authorized to read corresponding Binary attachment (see #362 )

cooperthompson:
I noticed this isn't marked as willfix yet.  I'm specifically interested in allowing the patient/Binary.read scope.  Is there an open question here about whether this will be fixed?  Or should this be marked as willfix?  If there is a question, should we split the patient/Binary.read back into #362 so we can fix that independent of this issue?

arscan:
We are not planning on expanding v1.x with any more tests, so I am closing this here.  Separately, v2.x should allow clients to request patient/Binary.read access, but please let us know if you are still seeing problems there @cooperthompson

PR #426, opened 2021-12-23
FI-1373 Add Attachment access test

yunwwang:
# Summary
This addresses GitHub Issue #374 

## Code changes
Update code generator to generate Attachment read test if the profile has an Attachment element with MustSuppot tag.

### DIFF
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -120,6 +120,7 @@ module Inferno
             add_terminology_bindings(profile_definition, new_sequence)
             add_search_param_descriptions(profile_definition, new_sequence)
             add_references(profile_definition, new_sequence)
+            add_attachments(profile_definition, new_sequence)
 
             metadata[:sequences] << new_sequence
           end
@@ -456,6 +457,11 @@ module Inferno
             profiles: ref_def['type'].first['targetProfile']
           }
         end
+      end
+
+      def add_attachments(profile_definition, sequence)
+        attachments = profile_definition['snapshot']['element'].select { |el| el['type'].present? && el['type'].first['code'] == 'Attachment' && el['mustSupport'] }
+        sequence[:attachments] = attachments.map { |el| el['path'].split('.', 2)[1] }
       end
 
       def add_mandatory_and_must_support_search_exclusions(metadata)
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -461,7 +461,7 @@ module Inferno
       end
 
       def add_attachments(profile_definition, sequence)
-        attachments = profile_definition['snapshot']['element'].select { |el| el['type'].present? && el['type'].first['code'] == 'Attachment' && el['mustSupport'] }
+        attachments = profile_definition['snapshot']['element'].select { |el| el.dig('base','path') == 'Attachment.url' && el['mustSupport'] }
         sequence[:attachments] = attachments.map { |el| el['path'].split('.', 2)[1] }
       end
 
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -461,7 +461,7 @@ module Inferno
       end
 
       def add_attachments(profile_definition, sequence)
-        attachments = profile_definition['snapshot']['element'].select { |el| el.dig('base', 'path') == 'Attachment.url' && el['mustSupport'] }
+        attachments = profile_definition['snapshot']['element'].select { |el| el['type'].present? && el['type'].first['code'] == 'Attachment' && el['mustSupport'] }
         sequence[:attachments] = attachments.map { |el| el['path'].split('.', 2)[1] }
       end
 
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -461,7 +461,7 @@ module Inferno
       end
 
       def add_attachments(profile_definition, sequence)
-        attachments = profile_definition['snapshot']['element'].select { |el| el.dig('base','path') == 'Attachment.url' && el['mustSupport'] }
+        attachments = profile_definition['snapshot']['element'].select { |el| el.dig('base', 'path') == 'Attachment.url' && el['mustSupport'] }
         sequence[:attachments] = attachments.map { |el| el['path'].split('.', 2)[1] }
       end
 
--- a/generator/uscore/metadata_extractor.rb
+++ b/generator/uscore/metadata_extractor.rb
@@ -85,6 +85,7 @@ module Inferno
           searches: [],
           search_param_descriptions: {},
           references: [],
+          attachments: [],
           must_supports: {
             extensions: [],
             slices: [],
--- a/generator/uscore/uscore_generator.rb
+++ b/generator/uscore/uscore_generator.rb
@@ -1028,16 +1028,23 @@ module Inferno
         test[:test_code] = %(
               #{skip_if_not_found_code(sequence)}
 
-              found_one_attachment = false
+              download_one_attachment = false
+              found_attachment_url = false
 
               #{resource_array}&.each do |resource|
-                if validate_attachment_resolutions(resource, [#{sequence[:attachments].map { |ele| "'#{ele}'" }.join(', ')}])
-                  found_one_attachment = true
-                  break
-                end
+              result = validate_attachment_resolutions(resource, [#{sequence[:attachments].map { |ele| "'#{ele}'" }.join(', ')}])
+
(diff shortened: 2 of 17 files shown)


lib/modules/uscore_v3.1.1/us_core_diagnosticreport_note_sequence.rb
```diff
@@ -729,6 +729,33 @@ def patient_ids
...
+        end
+
+        skip_if_not_found(resource_type: 'DiagnosticReport', delayed: false)
+
+        found_one_attachment = false
+
+        @diagnostic_report_ary&.values&.flatten&.each do |resource|
+          if validate_attachment_resolutions(resource, ['presentedForm'])
+            found_one_attachment = true
+            break
```
Jammjammjamm, line 753: The description says that this test attempts to read all attachments and fails if it is unable to read any, but this code indicates that it stops reading attachments once it successfully reads one, and as long as a single attachment can be read, all other can fail.

generator/uscore/metadata_extractor.rb
```diff
@@ -458,6 +460,11 @@ def add_references(profile_definition, sequence)
         end
       end
 
+      def add_attachments(profile_definition, sequence)
+        attachments = profile_definition['snapshot']['element'].select { |el| el['type'].present? && el['type'].first['code'] == 'Attachment' && el['mustSupport'] }
```
Jammjammjamm, line 464: I think this needs to check whether `attachment.url` is must support. It is for DocumentReference, but not for DiagnosticReport, so DiagnosticReport could only support inline attachments.

lib/app/views/onc_program.erb
```diff
@@ -648,6 +648,27 @@
...
+
+        <div class="form-group"
+          data-requiredby="<%= test_set.variable_required_by(:attachment_requires_token).map {|seq| seq.sequence_name}.join(',')%>"
+          data-prerequisite="attachment_requires_token"
+          id="op_attachment_requires_token"
+          >
+          <label style="display:block">Attachment URL Authorization
+            <small>Select SMART Bearer token if attachments are served from a location that accepts the same bearer token used for the SIngle Patient API.  
+            Select "Self-signed URL" if the attachments encode authorization information within the URL and bearer tokens are not required.  
+            Systems that use "Self-signed URLs" are expected to secure them in a manner consistent with SMART Bearer tokens.</small>
```
Jammjammjamm, line 660: > are expected to secure them in a manner consistent with SMART Bearer tokens.

What does this mean? Do we have a real requirement we could reference?

lib/modules/onc_program_module.yml
```diff
@@ -141,6 +141,7 @@ test_sets:
           These resources will not be tested for FHIR search support.
         lock_variables: 
           - token
+          - attachment_requires_token
```
Jammjammjamm, line 144: Why add this input if it is locked?

lib/app/views/onc_program.erb
```diff
@@ -648,6 +648,27 @@
...
               })%> 
 
+
+        <div class="form-group"
+          data-requiredby="<%= test_set.variable_required_by(:attachment_requires_token).map {|seq| seq.sequence_name}.join(',')%>"
+          data-prerequisite="attachment_requires_token"
+          id="op_attachment_requires_token"
+          >
+          <label style="display:block">Attachment URL Authorization
+            <small>Select SMART Bearer token if attachments are served from a location that accepts the same bearer token used for the SIngle Patient API.
```
Jammjammjamm, line 658:
```suggestion
            <small>Select SMART Bearer Token if attachments are served from a location that accepts the same bearer token used for the Single Patient API.  
```

lib/app/views/onc_program.erb
```diff
@@ -648,6 +648,27 @@
...
 
+
+        <div class="form-group"
+          data-requiredby="<%= test_set.variable_required_by(:attachment_requires_token).map {|seq| seq.sequence_name}.join(',')%>"
+          data-prerequisite="attachment_requires_token"
+          id="op_attachment_requires_token"
+          >
+          <label style="display:block">Attachment URL Authorization
+            <small>Select SMART Bearer token if attachments are served from a location that accepts the same bearer token used for the SIngle Patient API.  
+            Select "Self-signed URL" if the attachments encode authorization information within the URL and bearer tokens are not required.
```
Jammjammjamm, line 659: It seems inconsistent that `Self-signed URL` is quoted, but `SMART Bearer Token` isn't.

lib/modules/uscore_v3.1.1/us_core_diagnosticreport_note_sequence.rb
yunwwang: Will update the description to match the new testing logic.

generator/uscore/metadata_extractor.rb
yunwwang: It does:

{ |el| el['type'].present? && el['type'].first['code'] == 'Attachment' && el['mustSupport'] }

lib/app/views/onc_program.erb
yunwwang: This is derived from discussion of bulk data's requiresAccessToken element (see discussion [here](https://chat.fhir.org/#narrow/stream/179250-bulk-data/topic/301.20redirect)). The idea is that sever can tell client the attacment url is a regular url requires Bearer token or is a self-signed urls without Bearer token. Bulk Data v2 mentioned (similar requirement)[https://hl7.org/fhir/uv/bulkdata/export.html#attachments]

lib/modules/onc_program_module.yml
yunwwang: We think though server can support self-signed urls for attachment and we programed our testing to take this boolean option into logic. Since there is still debating if self-signed urls has the same effects as OAuth token, we decide to disable that at this release. If there is feedback from implementer community indicating preference of self-signed urls, it is easier to change this configuraiton than actual coding.

lib/modules/uscore_v3.1.1/us_core_diagnosticreport_note_sequence.rb
Jammjammjamm: Is only requiring one attachment to work the behavior we want?

lib/app/views/onc_program.erb
Jammjammjamm: It isn't clear to me from that discussion what "secure them in a manner consistent with SMART Bearer tokens" actually means.

generator/uscore/uscore_generator.rb
```diff
@@ -1004,6 +1010,38 @@ def create_references_resolved_test(sequence)
...
+      def create_attachment_test(sequence)
+        return unless sequence[:attachments].any?
+
+        test = {
+          tests_that: "At least one attachment within #{sequence[:resource]} resources can be read.",
+          index: sequence[:tests].length + 1,
+          link: 'https://www.hl7.org/fhir/datatypes.html#attachment',
+          description: %(
+            This test will attempt to read an attachment in the resources from the first search.
+            The test will pass if Inferno successfully download at least one attachment.
```
Jammjammjamm, line 1022:
```suggestion
            The test will pass if Inferno successfully downloads at least one attachment.
```

lib/modules/uscore_v3.1.1/us_core_diagnosticreport_note_sequence.rb
yunwwang: Yes. One is enough to demonstrate the server "support" attachment download

lib/app/views/onc_program.erb
yunwwang: My understanding is the same way server can handle access token, like server can revoke, expire and renew an access token. Should (or could) server do the same on an self-signed url?

yunwwang: The actual expectation for self-signed url is still up for discussion as mentioned above. So we mentioned there is an option to use self-signed url *but Inferno Program testing does NOT allow that option at this time*

Jammjammjamm (approved):

generator/uscore/metadata_extractor.rb
yunwwang: Reverse the changes since attachment.url or attachment.data is must supported.

arscan:
I"m a bit concerned that the language is confusing because:
1) I"m not sure 'self-signed URL' is an industry term here
2) This implies that the user can change this ("Select ..."), when they can't because we've locked it (for now). 

How about this:

Screen Shot 2022-01-13 at 3 01 47 PM

arscan:
This will not be implemented in v1.x, we will consider adding this to v2.x separately.

